PR #7, opened 2015-08-21
ID lookup

kwm4385:
Added ID lookup api endpoints for the UI so we don't need ugly urls.


kwm4385:
Hold on this until I merge in @jhaber 's PR.

kwm4385:
This should be ready to go @tpetr @jhaber

BlazarUI/app/scripts/models/BranchDefinition.js
```diff
@@ -0,0 +1,16 @@
...
+class BranchDefinition extends Model {
+
+  constructor(gitInfo) {
+    super()
+    this.gitInfo = gitInfo;
+  }
+
+  url() {
+    let gitInfo = this.gitInfo;
+    return `${config.apiRoot}/branch/lookup?host=${gitInfo.url}&organization=${gitInfo.org}&repository=${gitInfo.repo}&branch=${gitInfo.branch}`;
```
tpetr: need to keep these URLs consistent -- can you remove the trailing slash before branch?

kwm4385: Shall I remove all instances of this and have the convention be the api root contains the trailing slash?

PR #8, opened 2015-08-24
ID lookup

jhaber:
Move functionality into BranchResource (new) and BuildHistoryResource


kwm4385:
Merging this to the original branch.

PR #9, opened 2015-08-24
Convert to conventions

kwm4385:
no trailing slash in roots
static root independent from appRoot
no conditional on appRoot template

Tested working in local (gulp), BlazarService, and docker.


kwm4385:
@tpetr

PR #10, opened 2015-08-26
ZooKeeper

jhaber:
Update EventBus to be backed by ZooKeeper so we don't lose data if a server dies


jhaber:
Seems to work, just need to update docker setup to pass zookeeper configuration on startup

PR #12, opened 2015-08-27
Make module discovery extensible

jhaber:


BlazarService/src/main/java/com/hubspot/blazar/util/ModuleDiscovery.java
```diff
@@ -1,69 +1,13 @@
...
-
-    return Preconditions.checkNotNull(gitHubByHost.get(host), "No GitHub found for host " + host);
-  }
-
-  private static boolean isPom(String path) {
-    return "pom.xml".equals(path) || path.endsWith("/pom.xml");
-  }
+public interface ModuleDiscovery {
+  boolean shouldRediscover(GitInfo gitInfo, PushEvent pushEvent) throws IOException;
+  Set<Module> discover(GitInfo gitInfo) throws IOException;
```
jhaber: @also this is the interface you would need to implement (you can extend `AbstractModuleDiscovery` to get some helper methods and you can look at `MavenModuleDiscovery` for an example impl)

PR #15, opened 2015-09-05
(WIP) Module focus instead of repo/branch

benrlodge:
This changes the previous organization of the sidebar (and starred dashboard items) from a repo/branch focus to a module focus. The only noticeable change should be the sidebar and the dashboard page (though there was some refactoring throughout the UI when it affected this new setup), everything else should still function the same. The one feature not updated to take into account this new setup is the notifications, I'd like to merge this and tackle the build log first so the whole ui is functioning, then we can add that back in.

Sidebar
 [x] - organize and list modules instead of module by repo/branch
 [x] - update star functionality to use module instead of repo/branch

Dashboard
 [x] - change star setup for new modules setup
 [x] - one list (not inactive/active)
 [ ] - change notification setup


BlazarUI/app/scripts/components/sidebar/BuildsSidebar.jsx
```diff
@@ -1,43 +1,49 @@
...
 import SidebarFilter from './SidebarFilter.jsx';
 import fuzzy from 'fuzzy';
-import {bindAll, filter, contains} from 'underscore';
+import {bindAll, filter, contains, has} from 'underscore';
 import SectionLoader from '../shared/SectionLoader.jsx';
-import LazyRender from '../shared/LazyRender.jsx';
 import Helpers from '../ComponentHelpers';
 import MutedMessage from '../shared/MutedMessage.jsx';
+import SidebarItem from '../sidebar/SidebarItem.jsx';
+import Immutable from 'immutable';
```
also: There's no `package.json` dependency on `immutable`, but it looks like it isn't actually used here.

PR #17, opened 2015-09-11
Build log page updates

benrlodge:
Adds
 - basic tailing
 - updated styles

round 2 will include
 - smoother transitions with new log lines
 - links to share log lines
 - explore collapsable lines
 - general ux stuff - e.g. scroll to bottom, scroll to top

### DIFF
--- a/BlazarUI/app/scripts/actions/buildActions.js
+++ b/BlazarUI/app/scripts/actions/buildActions.js
@@ -1,6 +1,6 @@
 /*global BuildHistoryActions*/
 import Reflux from 'reflux';
-import _ from 'underscore';
+import {find} from 'underscore';
 
 import Build from '../models/Build';
 import Log from '../models/Log';
@@ -34,7 +34,7 @@ function getModule() {
   const modulesPromise = branchModules.fetch();
 
   modulesPromise.done( () => {
-    gitInfo.moduleId = _.find(branchModules.data, (m) => {
+    gitInfo.moduleId = find(branchModules.data, (m) => {
       return m.name === gitInfo.module;
     }).id;
     getBuild();
@@ -57,7 +57,6 @@ function getBranchId() {
     });
 }
 
-
 function getBuild() {
   const build = new Build(gitInfo);
   const buildPromise = build.fetch();
@@ -74,26 +73,23 @@ function getBuild() {
     const log = new Log(build.data.build.id);
     const logPromise = log.fetch();
 
-    logPromise.always( (logData) => {
-      if (!logData) {
-        logData.data = '';
-      }
+    logPromise.always( (data, textStatus, jqxhr) => {
 
       BuildActions.loadBuildSuccess({
         build: build.data,
-        log: logData.data
+        log: log.formatLog(jqxhr)
       });
 
     });
 
     logPromise.error( () => {
-      BuildActions.loadBuildError('Error retrieving build log');
+      BuildActions.loadBuildError("<p class='x-y-roomy'>Error retrieving build log");
     });
 
   });
 
   buildPromise.error( () => {
-    BuildActions.loadBuildError('Error retrieving build');
+    BuildActions.loadBuildError("<p class='x-y-roomy'>Error retrieving build");
   });
 
 }
--- a/BlazarUI/app/scripts/components/build/Build.jsx
+++ b/BlazarUI/app/scripts/components/build/Build.jsx
@@ -60,6 +60,7 @@ class Build extends Component {
           <UIGridItem size={12}>
             <BuildLog
               log={this.props.log}
+              isBuilding={build.state === 'IN_PROGRESS'}
               loading={this.props.loading}
             />
           </UIGridItem>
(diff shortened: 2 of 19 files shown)


BlazarUI/app/scripts/actions/buildActions.js
```diff
@@ -34,7 +34,7 @@ function getModule() {
   const modulesPromise = branchModules.fetch();
 
   modulesPromise.done( () => {
-    gitInfo.moduleId = _.find(branchModules.data, (m) => {
+    gitInfo.moduleId = find(branchModules.data, (m) => {
```
gusvargas: Do we want to target IE?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

benrlodge: I'm actually using underscore's find, I was just destructuring here.

BlazarUI/app/scripts/components/build/BuildCommits.jsx
```diff
@@ -19,8 +20,9 @@ class BuildCommits extends Component {
       return <SectionLoader />;
     }
 
+    const inflection = commitInfo.newCommits.length !== 1 ? 's' : '';
     const header = (
-      <span> ({commitInfo.newCommits.length}) new commits since previous build </span>
+      <span> <span className='badge'>{commitInfo.newCommits.length}</span> new commit{inflection} since previous build </span>
```
gusvargas: For readability you should nest inner tags instead of putting them inline.

BlazarUI/app/scripts/actions/buildActions.js
gusvargas: I know was just wondering if we need to use `underscore` for functions that already exist natively on object prototypes in modern browsers.

BlazarUI/app/scripts/models/Log.js
```diff
@@ -14,8 +13,27 @@ class Log extends Model {
...
+      return "<p class='x-y-roomy'>Error loading build log. Please check your console for more detail.</p>";
+    }
+
+    const lines = jqxhr.responseJSON.data.split('\n');
+
+    lines.forEach( (line) => {
+      cache.push(`<p class='log-line'>${line}</p>`);
+    });
+
+    return cache.join('\n');
```
gusvargas: It's more idiomatic to express this using `Array.prototype.map`

```javascript
const NEW_LINE = '\n';
return jqxhr.responseJSON.data.split(NEW_LINE).map((line) => {
  return `<p class='log-line'>${line}</p>`;
}).join(NEW_LINE);
```

BlazarUI/app/scripts/actions/buildActions.js
benrlodge: ah gotcha. yea I'm guessing we don't need to be too worried about IE given Blazar's typical user.

BlazarUI/app/scripts/models/Log.js
benrlodge: cool thanks

BlazarUI/app/scripts/components/build/Build.jsx
```diff
@@ -60,6 +60,7 @@ class Build extends Component {
           <UIGridItem size={12}>
             <BuildLog
               log={this.props.log}
+              isBuilding={build.state === 'IN_PROGRESS'}
```
gusvargas: We should make a `BuildStates` module that enumerates all the build states. Then you can do something like:

```javascript
// /constats/BuildStates.js
export default keyMirror({
  IN_PROGRESS: null,
  // etc

// Build.jsx
// ... snip ...
    isBuilding={build.state === BuildStates.IN_PROGRESS}
// ... snip ...
});
```

It's nice to have a central place with all build states within the context of the UI code instead of having to find all `build.state` usages to grok all possible values.

More benefits [here](http://stackoverflow.com/a/31347918)

benrlodge: nice, i'll take a look at that

BlazarUI/app/scripts/models/Log.js
gusvargas: On second thought, I have a question about this -- why the newline character between `p` tags? `p` elements are `display: block` so they should render on newlines naturally. I don't think that the `\n` is actually doing anything. Is there another use of this output that I may not be considering? I'm assuming it's only being used as inner html.

benrlodge: Let me know if I didn't hear you correctly, I'm not sure if I get what you mean when you say `between p tags` - The `responseJSON.data` returns one long string, so I' using the `\n` to split it, to add the `log-line` class for some css formatting, but I'm also planning on adding links to line numbers in here (like how you can on github, and recently in singularity), so I'll need a place to add that. FWIW, this was a quick pass to get the log showing, I was going to think about it a bit more, but wanted to get something out, so if you had any thoughts on the setup in general, I'm happy to hear.

gusvargas: Yeah I understand the initial splitting by `\n` but I'm not sure why you're doing a `.join('\n')` after creating the `p` elements. You're outputing:

```html
<p class='log-line'>Foo</p>\n
<p class='log-line'>Foo</p>\n
<p class='log-line'>Foo</p>\n
<p class='log-line'>Foo</p>\n
```

Maybe I'm missing something?

benrlodge: I see what was going on.  I had the `log-lines` set as inline-block, with the container as a pre tag, the join with `\n` was fixing the issue of them being displayed inline. Dropped the inline-block and it's all set. Thanks for noticing that.

PR #18, opened 2015-09-15
Add build now button

kwm4385:


BlazarUI/app/scripts/components/module/BuildButton.jsx
```diff
@@ -0,0 +1,52 @@
...
+  constructor(props) {
+    super(props);
+    this.state = {loading: false, textStatus: ''};
+  }
+
+  build() {
+    this.setState({
+      loading: true
+    });
+    $.ajax({
```
benrlodge: We shouldn't have ajax requests in components, we should put them in an action

BlazarUI/app/scripts/components/module/BuildButton.jsx
```diff
@@ -0,0 +1,52 @@
...
+      this.setState({
+        loading: false,
+        textStatus: jqXHR
+      });
+    });
+  }
+
+  render() {
+    if (this.state.loading) {
+      return <Button bsStyle="primary" disabled><Icon type="fa" name="circle-o-notch" classNames="fa-spin" /> Build Now</Button>;
```
benrlodge: I added a new prop to the [Icon component](https://github.com/HubSpot/Blazar/blob/master/BlazarUI/app/scripts/components/shared/Icon.jsx#L11) called `for` with a mapping in the [constants file](https://github.com/HubSpot/Blazar/blob/master/BlazarUI/app/scripts/components/constants.js#L24) so we don't need to worry about the exact names of the fa and octicon icons.  I still need to go back to update old Icons, but moving forward let's use this.

benrlodge: I may have done some one liners in the past, but moving forward let's try to always return markup wrapped in parents and indented to improve readability. e.g. -
```javascript
return (
   <Button bsStyle="primary" disabled>
     <Icon type="fa" name="circle-o-notch" classNames="fa-spin" /> Build Now
  </Button>
);
```

kwm4385:
:+1:

benrlodge:
better, but we can do more! 

Let's keep all data fetching/loading/communication with the actions in our container components (eg in this case moduleContainer), and pass the data and functions to call the data down as props. We shoudl keep all promise related stuff in the actions (we shouldn't have stuff like .then(() => { in the view components), views should just consume props as data/functions, and manipulate state. Instead have your `build()` method call a function that is passed down as props from the moduleContainer. Take a look at the starring feature that I refactored (look at `persistStarChange` in `/components/sidebar/BuildsSidebarContainer.jsx`, and see how `starActions` communicates with the `starStore`, passing down the data to the container, and then to the the final component. So your textStatus and loading info should come as props from the container.

BlazarUI/app/scripts/components/module/BuildButton.jsx
```diff
@@ -0,0 +1,55 @@
...
+    super(props);
+    this.state = {loading: false, textStatus: ''};
+  }
+
+  build() {
+    this.setState({
+      loading: true
+    });
+
+    BuildsActions.triggerBuild(this.props.moduleId).then(() => {
```
benrlodge: move to the action/store

BlazarUI/app/scripts/components/module/BuildButton.jsx
```diff
@@ -0,0 +1,55 @@
...
+    if (this.state.loading) {
+      return (
+        <Button bsStyle="primary" disabled>
+          <Icon for="spinner" /> Build Now
+        </Button>
+      );
+    } else {
+      return (
+        <div>
+          <span>{this.state.textStatus} </span>
```
benrlodge: `this.state.textStatus` should come down as props

BlazarUI/app/scripts/components/module/BuildButton.jsx
```diff
@@ -0,0 +1,36 @@
...
+  render() {
+    if (this.props.buildTriggering) {
+      return (
+        <Button bsStyle="primary" disabled>
+          <Icon for="spinner" /> Build Now
+        </Button>
+      );
+    } else {
+      return (
+        <Button bsStyle="primary" onClick={this.build.bind(this)}> Build Now</Button>
```
benrlodge: May be more of a preference, but for consistency's sake, I've been trying to keep the bindings of these types of functions in the constructor and using underscore's bindAll. It just makes the markup a little cleaner and organized IMO. (like -> https://github.com/HubSpot/Blazar/blob/c59cef468a37f71a954f18f6237b76f9bed6b7ae/BlazarUI/app/scripts/components/sidebar/BuildsSidebar.jsx#L19)

BlazarUI/app/scripts/components/module/ModuleContainer.jsx
```diff
@@ -11,13 +11,16 @@ class ModuleContainer extends Component {
...
+      loading: true,
+      buildTriggeringDone: true,
+      buildTriggeringError: ''
     };
   }
 
   componentDidMount() {
     this.unsubscribe = BuildHistoryStore.listen(this.onStatusChange.bind(this));
     BuildHistoryActions.loadBuildHistory(this.props.params);
+    this.buildTriggered = BuildHistoryStore.listen(this.onStatusChange.bind(this));
```
benrlodge: this is redundant as we're already listening to the store on line 21

BlazarUI/app/scripts/components/module/ModuleContainer.jsx
```diff
@@ -40,6 +47,8 @@ class ModuleContainer extends Component {
           params={this.props.params}
           buildHistory={this.state.buildHistory}
           loading={this.state.loading}
+          triggerBuild={this.triggerBuild.bind(this)}
```
benrlodge: same as comment ^, move binding to constructor

BlazarUI/app/scripts/components/module/BuildButton.jsx
```diff
@@ -0,0 +1,37 @@
...
+    super(props);
+    bindAll(this, 'build');
+  }
+
+  build() {
+    this.props.triggerBuild();
+  }
+
+  render() {
+    if (this.props.buildTriggering) {
```
benrlodge: when i try to build it flashes with the spinner and the disabled state, but then goes back to the normal state almost immediately. Also, we may need to get the current build state to determine if we want this disabled or not, rather than the buildTrigger. Say you land on this page and a build is in progress, the button should be disabled. or alternatively, some notification should show up saying a build is in progress and prevent the build.

kwm4385: You can still trigger a build while one is in progress, it will simply be queued for when it's done. The button is only disabled while the request is pending, which normally will be pretty fast.

PR #19, opened 2015-09-15
Shrink header/breadcrumbs when scrolling

kwm4385:


BlazarUI/app/scripts/components/shared/PageHeader.jsx
```diff
@@ -1,6 +1,19 @@
 import React, {Component, PropTypes} from 'react';
+import $ from 'jquery';
```
benrlodge: i'd like as much as we can to avoid jquery and instead use react where we can, e.g. -> https://facebook.github.io/react/docs/animation.html

gusvargas: https://github.com/captivationsoftware/react-sticky is this useful here?

:+1: to not falling back to jQuery when the animation stuff isn't obvious w/ React.

BlazarUI/app/scripts/components/dashboard/Dashboard.jsx
```diff
@@ -41,6 +41,7 @@ class Dashboard extends Component {
     return (
       <div>
         <PageHeader>
+          <br/>
```
gusvargas: Use `margin-bottom` on the `PageHeader` instead of using `<br />` tags.

BlazarUI/app/scripts/components/shared/PageHeader.jsx
kwm4385: @benrlodge The react animation docs seem to only cover the use case of animating components as the enter or leave the DOM. Does it work if we want to modify an existing element?

benrlodge: i've honestly done little with react animations, but take a look at -> http://unitstep.net/blog/2015/03/03/using-react-animations-to-transition-between-ui-states this might do the trick

kwm4385:
jQuery has been terminated

benrlodge:
still working on this? I'm getting `Cannot read property 'scrollTop' of undefined` error

kwm4385:
Weird, it's working for me. Mind posting the entire stack trace?

kwm4385:
Just kidding. Try it now, should be working.

BlazarUI/app/stylus/type.styl
```diff
@@ -47,4 +47,7 @@ td
 
 .notfound__heading
   font-weight 100
-  margin-bottom 20px
\ No newline at end of file
+  margin-bottom 20px
+
+.headline-no-breadcrumbs
```
benrlodge: move into headline component's styl file since this is specifically addressing that component. Also, try not to resort to using !important when it's not necessary, it causes much more trouble than good in the long run, just try to be [more specific](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity) - e.g.

```stylus
.headline
  font-size 26px
  margin-top 8px
  margin-bottom 0
  padding 5px 0 10px 0

  &.headline-no-breadcrumbs
    margin-top 35px
```

BlazarUI/app/scripts/components/shared/PageHeader.jsx
```diff
@@ -1,10 +1,38 @@
...
+    window.onscroll = this.handleScroll.bind(this);
+  }
+
+  componentWillUnmount() {
+    window.onscroll = null;
+  }
+
+  handleScroll(event) {
+    let position = window.pageYOffset;
+    if (position > 10) {
```
benrlodge: there's a lot of unnecessary calls to set state and render here. add a state property that keeps track of if we want the page header to be small, then you can do something like `if (position > 10 && !this.state.isPageHeaderSmall)`.

kwm4385:
Good suggestions

BlazarUI/app/scripts/components/shared/PageHeader.jsx
```diff
@@ -1,10 +1,41 @@
 import React, {Component, PropTypes} from 'react';
 
-class PageHeader extends Component{
+class PageHeader extends Component {
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      headerClass: '',
+      isSmall: false
```
benrlodge: lets simplify the way we handle the state, I think we can get away with one state property and adjust the render method to show the right class

PR #21, opened 2015-09-16
Make sidebar collapsable

kwm4385:


BlazarUI/app/scripts/components/sidebar/BuildsSidebarContainer.jsx
```diff
@@ -56,17 +57,30 @@ class BuildsSidebarContainer extends Component {
...
     return (
-      <Sidebar>
-        <BuildsSidebar
-          builds={this.state.builds}
-          stars={this.state.stars}
-          loading={this.state.loading} 
-          persistStarChange={this.persistStarChange} />
-      </Sidebar>
+      <div>
+        {miniLogo}
```
gusvargas: Rather than putting conditional logic in `render`, use a helper like `this.renderMiniLogo()`

```javascript
renderMiniLogo() {
  if (!this.props.isCollpased) {
    return null;
  }

  return (
    <SidebarLogo mini={true} />
  );
}
```

BlazarUI/app/scripts/components/sidebar/Sidebar.jsx
```diff
@@ -1,14 +1,25 @@
...
+    const sidebarClasses = this.props.isCollapsed ? 'sidebar-collapsed' : '';
+
     let headline;
     if (this.props.headline) {
       headline = <h2 className='sidebar__headline'>{this.props.headline}</h2>;
     }
     return (
-      <div className='sidebar'>
+      <div className={`sidebar ${sidebarClasses}`}>
+        {collapseButton}
```
gusvargas: Same as above, keep `render` as declarative as possible. Adding conditional logic in here become a slippery slope. Looks like this component could make use of `renderCollapseButton` and `renderHeadline`.

BlazarUI/app/scripts/components/sidebar/SidebarLogo.jsx
```diff
@@ -5,16 +5,28 @@ import Image from '../shared/Image.jsx';
...
+    } else {
+      imgPath = `${window.config.staticRoot}/images/blazar-logo.png`;
+      image = <Image classNames="sidebar__logo-image" src={imgPath} />;
+    }
+
     return (
       <div className="sidebar__logo">
         <Link to="dashboard">
-          <Image classNames="sidebar__logo-image" src={imgPath} />
+        {image}
```
gusvargas: Same here. Also I think you meant to indent this line.

```javascript
getLogoImageProps() {
  if (this.props.mini) {
    return {
      src: `${window.config.staticRoot}/images/icon.png`,
      width: 35,
      height: 35
    };
  }

  return {
    src: `${window.config.staticRoot}/images/blazar-logo.png`
  };
},

renderSidebarLogo() {
  return (
    <Image
      className='sidebar__logo-image'
      {...getLogoImageProps()} />
  );
},

render() {
  return (
    // ... snip ...
      {this.renderSidebarLogoImage()}
   // ... snip ...
  );
}
```

BlazarUI/app/scripts/pages/app.jsx
```diff
@@ -4,12 +4,15 @@ import BuildsSidebarContainer from '../components/sidebar/BuildsSidebarContainer
...
 import Input from 'react-bootstrap/lib/Input';
+import {bindAll} from 'underscore';
 
 class App extends Component {
 
   constructor() {
+    bindAll(this, 'collapseSidebar');
     this.state = {
-      showModal: (!window.config.apiRoot)
+      showModal: (!window.config.apiRoot),
```
gusvargas: The parens aren't necessary.

BlazarUI/app/scripts/pages/app.jsx
```diff
@@ -4,12 +4,15 @@ import BuildsSidebarContainer from '../components/sidebar/BuildsSidebarContainer
...
+      sidebarCollapsed: localStorage.sidebarCollapsed === 'true' ? true : false
```
gusvargas: You don't need the ternary operator, it can be rewritten as:

```javascript
  sidebarCollapsed: localStorage.sidebarCollpased === 'true'
```

Also, @benrlodge there are a couple of concerning bits in this module:

1. It's accessing and modifying globals directly (`localStorage` and `window.config.apiRoot`). These globals are also used to determine the UI state. A react component ideally should be a pure function of `state`, `props`, and `context`.
2. For some reason it's modifying `this.state` direct then calling `this.forceUpdate()` instead of just using `this.setState`

Suggested solutions:

1. Treat reading from globals the same way you would reading from an API. I'm not familiar with the Flux framework you're using but if you have a model for an API then you should have a model for `localStorage` that this module can subscribe to. This will also make your components easier to unit test since you won't have to stub `localStorage` and `window` reads.
2. Use `this.setState`

benrlodge: thanks @gusvargas - I noticed these as well and was planning on reviewing with kevin

BlazarUI/app/scripts/components/sidebar/BuildsSidebarContainer.jsx
benrlodge: Is there a reason we need to return null? I know I've read you can/should return null for an empty component, but I haven't seen anything that said you can return undefined. That said, it appear to work returning undefined? 

```javascript
renderMiniLogo() {
  if (this.props.isCollpased) {
    return (
      <SidebarLogo mini={true} />
    );
}
```

kwm4385:
Did some cleanup of those concerns @gusvargas, we'll be working to refactor apiModal and localStorage use in another issue.

BlazarUI/app/scripts/components/sidebar/BuildsSidebarContainer.jsx
gusvargas: You're right, `undefined` totally works but it's more explicit to check for the empty case and return `null` instead of letting implicit `undefined` returns do it for you. Definitely more of a stylistic thing but explicit > implicit.

benrlodge:
I was giving this some more thought, I'd like to try to simplify this, I think we can avoid using local storage, let's discuss later.

benrlodge:
the  layout has changed substantially since this update so I'm closing this for now and will work on a new pr when we get this.

PR #30, opened 2015-10-05
only request build updates after initial request

benrlodge:
Previously all builds were fetched during each request (a whopping ~2mb every 5 seconds), this update passes a ?since parameter which only receives builds that have updated since the last request (only ~1-4kb), and updates the builds store with any of these updates. 

@tpetr - right now I'm just storing the builds it in memory - after thinking about it I'm not sure there is a great benefit keeping it in local storage (and could potentially be less useful given local storage is synchronous and it would be hit often) since it's only the first request which is large (eg 2mbs), then its typically quite small.

//cc @jhaber

### DIFF
--- a/BlazarUI/app/scripts/collections/Builds.js
+++ b/BlazarUI/app/scripts/collections/Builds.js
@@ -4,8 +4,17 @@ import BaseCollection from './BaseCollection';
 
 class Builds extends BaseCollection {
 
+  constructor() {
+    this.cache = null;
+    this.updatedTimestamp = '';
+  }
+
+  parse() {
+    super.parse();
+  }
+
   url() {
-    return `${config.apiRoot}/build/states?property=!lastBuild.commitInfo&property=!inProgressBuild.commitInfo&property=!pendingBuild.commitInfo`;
+    return `${config.apiRoot}/build/states?property=!lastBuild.commitInfo&property=!inProgressBuild.commitInfo&property=!pendingBuild.commitInfo&since=${this.updatedTimestamp}`;
   }
 
   hasBuildState() {
@@ -53,7 +62,7 @@ class Builds extends BaseCollection {
   }
 
   getBranchModules(branchInfo) {
-    const modules = findWhere(this.groupBuildsByRepo(), {
+    const modules = findWhere(this.groupByRepo(), {
       organization: branchInfo.org,
       repository: branchInfo.repo,
       branch: branchInfo.branch
@@ -63,7 +72,7 @@ class Builds extends BaseCollection {
   }
 
   getBranchesByRepo(repoInfo) {
-    const groupedBuilds = this.groupBuildsByRepo();
+    const groupedBuilds = this.groupByRepo();
     let branches = groupedBuilds.filter((repo) => {
       return (repo.organization === repoInfo.org) && (repo.repository === repoInfo.repo);
     });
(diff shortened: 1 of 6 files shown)


BlazarUI/app/scripts/collections/Builds.js
```diff
@@ -4,8 +4,16 @@ import BaseCollection from './BaseCollection';
 
 class Builds extends BaseCollection {
 
+  constructor() {
+    this.updatedTimestamp = '';
```
jhaber: Initialize to 0?

benrlodge: cool, if 0 gets the start, better to be explicit.

BlazarUI/app/scripts/stores/buildsStore.js
```diff
@@ -1,5 +1,29 @@
 import Reflux from 'reflux';
 import BuildsActions from '../actions/buildsActions';
+import {pluck} from 'underscore';
+
+
+function getBuildsIds(builds) {
+  return pluck(pluck(builds, 'module'), 'id');
```
also: This shouldn't need two plucks. Something like

```javascript
builds.map(({module: {id}}) => id)
// or
builds.map(build => build.module.id)
```

BlazarUI/app/scripts/stores/buildsStore.js
```diff
@@ -1,5 +1,29 @@
...
+}
+
+function updateBuilds(latest, builds) {
+
+  const latestIds = getBuildsIds(latest);
+
+  for (let i = 0, len = builds.length; i < len; i++) {
+    for (let g = 0, len = latestIds.length; g < len; g++) {
+      if (latestIds[g] === builds[i].module.id) {
```
also: Why exclude new module ids?

BlazarUI/app/scripts/stores/buildsStore.js
```diff
@@ -1,5 +1,29 @@
 import Reflux from 'reflux';
 import BuildsActions from '../actions/buildsActions';
+import {pluck} from 'underscore';
+
+
+function getBuildsIds(builds) {
+  return pluck(pluck(builds, 'module'), 'id');
+}
+
+function updateBuilds(latest, builds) {
```
also: Could this be implemented with `builds

BlazarUI/app/scripts/stores/buildsStore.js
benrlodge: cool thanks

BlazarUI/app/scripts/stores/buildsStore.js
benrlodge: not following your question?

BlazarUI/app/scripts/stores/buildsStore.js
benrlodge: yea looking at this again, that getBuildsIds is completely unecessary

benrlodge:
there's a bug in here when switching between in-progress builds i'm looking into.

BlazarUI/app/scripts/stores/buildsStore.js
also: Only modules already in builds are updated, so if a new module is created it won't be added to `builds` this way, but it would when fetching the complete list.

benrlodge: ah, i see. thanks for picking that up.

PR #38, opened 2015-10-20
WIP: move buildpack config into BlazarConfiguration

tpetr:
this builds but i haven't tested it at all yet -- wanted to get some :eyes: and :thought_balloon: first

/cc @jonathanwgoodwin @jhaber @also

### DIFF
--- a/BlazarService/src/main/java/com/hubspot/blazar/config/BlazarConfiguration.java
+++ b/BlazarService/src/main/java/com/hubspot/blazar/config/BlazarConfiguration.java
@@ -1,5 +1,6 @@
 package com.hubspot.blazar.config;
 
+import java.util.HashMap;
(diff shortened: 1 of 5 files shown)


BlazarService/src/main/java/com/hubspot/blazar/config/BlazarConfiguration.java
```diff
@@ -37,6 +38,11 @@
   @JsonProperty("ui")
   private UiConfiguration uiConfiguration = new UiConfiguration();
 
+  @Valid
```
jhaber: I don't think `@Valid` works for maps

jonathanwgoodwin:
I think 238f7a9aa2d2db9b104e9251c2842f0a75a5bb95 closes this.

PR #42, opened 2015-10-28
Feedback form

kwm4385:
screen shot 2015-10-28 at 7 08 19 pm
 /cc @jonathanwgoodwin @jhaber


benrlodge:
Looks good @kwm4385 thanks for putting this together. I just made a few tweaks, moved the xhr to actions/store, allowed the form to be submitted again after a submission, and tweaked the styles a little bit. Once we get the slack api back up and working i'll merge this in.

benrlodge:
Unfortunately, it's not an accurate way to determine a browser vendor and version. While there are some workarounds, for the purpose and audience of this form, it will be easier/more accurate to just ask people to tell us what they are using. I couldn't find a great write-up, but here's a start -  https://lukasa.co.uk/2013/04/User_Agent_Strings/

PR #46, opened 2016-01-04
More flexible build command structure for new executor, this will break the old executor

jonathanwgoodwin:
This change will break the old executor script we were using. But it makes way for the new one.
@tpetr @jhaber

### DIFF
--- a/BlazarBase/src/main/java/com/hubspot/blazar/base/BuildCommand.java
+++ b/BlazarBase/src/main/java/com/hubspot/blazar/base/BuildCommand.java
@@ -28,7 +28,11 @@ public class BuildCommand {
                       @JsonProperty("returnCodes") Set<Integer> returnCodes,
                       @JsonProperty("env") Map<String, String> commandSpecificEnvironment) {
     this.executable = Objects.firstNonNull(executable, DEFAULT_CMD);
-    this.args = Objects.firstNonNull(args, Lists.newArrayList(DEFAULT_ARG));
+    if (executable == null) {
+      this.args = Lists.asList(DEFAULT_ARG, args.toArray(new String[args.size()]));
+    } else {
+      this.args = Objects.firstNonNull(args, Lists.newArrayList(DEFAULT_ARG));
+    }
     this.returnCodes = Objects.firstNonNull(returnCodes, DEFAULT_RETURN_CODES);
     this.commandSpecificEnvironment = Objects.firstNonNull(commandSpecificEnvironment, new HashMap<String, String>());
   }
--- a/BlazarBase/src/main/java/com/hubspot/blazar/base/BuildCommand.java
+++ b/BlazarBase/src/main/java/com/hubspot/blazar/base/BuildCommand.java
@@ -1,42 +1,41 @@
 package com.hubspot.blazar.base;
 
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.common.base.Objects;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Lists;
 
 public class BuildCommand {
-  private static String DEFAULT_CMD = "/bin/bash";
-  private static String DEFAULT_ARG = "-c";
+  private static final String DEFAULT_CMD = "/bin/bash";
+  private static final String DEFAULT_ARG = "-c";
+  public static final Set<Integer> DEFAULT_RETURN_CODES = ImmutableSet.of(0);
 
   private final String executable;
   private final List<String> args;
-  private final List<Integer> returnCodes;
+  private final Set<Integer> returnCodes;
   private final Map<String, String> commandSpecificEnvironment;
 
   @JsonCreator
   public BuildCommand(@JsonProperty("executable") String executable,
                       @JsonProperty("args") List<String> args,
-                      @JsonProperty("returnCodes") List<Integer> returnCodes,
+                      @JsonProperty("returnCodes") Set<Integer> returnCodes,
                       @JsonProperty("env") Map<String, String> commandSpecificEnvironment) {
     this.executable = Objects.firstNonNull(executable, DEFAULT_CMD);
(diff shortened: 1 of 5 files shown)


jhaber:
Do you have an example of where this would be needed?

jonathanwgoodwin:
Yeah I'll slack you a link

BlazarBase/src/main/java/com/hubspot/blazar/base/BuildCommand.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.Map;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.base.Objects;
+import com.google.common.collect.Lists;
+
+public class BuildCommand {
+  private static String DEFAULT_CMD = "/bin/bash";
+  private static String DEFAULT_ARG = "-c";
```
tpetr: these should be `final`

BlazarBase/src/main/java/com/hubspot/blazar/base/BuildCommand.java
```diff
@@ -0,0 +1,61 @@
...
+
+  private final String executable;
+  private final List<String> args;
+  private final List<Integer> returnCodes;
+  private final Map<String, String> commandSpecificEnvironment;
```
tpetr: seems a little too verbose -- `env` or `environment` or even `environmentVariables` is probably fine

BlazarBase/src/main/java/com/hubspot/blazar/base/BuildCommand.java
```diff
@@ -0,0 +1,61 @@
...
+
+  @JsonCreator
+  public BuildCommand(@JsonProperty("executable") String executable,
+                      @JsonProperty("args") List<String> args,
+                      @JsonProperty("returnCodes") List<Integer> returnCodes,
+                      @JsonProperty("env") Map<String, String> commandSpecificEnvironment) {
+    this.executable = Objects.firstNonNull(executable, DEFAULT_CMD);
+    this.args = Objects.firstNonNull(args, Lists.newArrayList(DEFAULT_ARG));
```
tpetr: not sure if we'd want `-c` to be the default argument here

BlazarBase/src/main/java/com/hubspot/blazar/base/BuildCommand.java
jonathanwgoodwin: It seems verbose here, but I'd like to keep it clearly separate from the other environments that we have floating around, such as the environment for this section of the build. And the environment for the whole build.

BlazarBase/src/main/java/com/hubspot/blazar/base/BuildCommand.java
```diff
@@ -0,0 +1,61 @@
...
+    this.returnCodes = Objects.firstNonNull(returnCodes, Lists.newArrayList(0));
```
tpetr: instead of making a new `[0]` with each instantiation, create a `public static final Set<Integer> DEFAULT_RETURN_CODES = ImmutableSet.of(0)`

BlazarBase/src/main/java/com/hubspot/blazar/base/BuildCommand.java
```diff
@@ -0,0 +1,61 @@
...
+    this.commandSpecificEnvironment = Objects.firstNonNull(commandSpecificEnvironment, new HashMap<String, String>());
```
tpetr: `Collections.emptyMap()` is preferable because it uses a hardcoded empty map, instead of instantiating a new one each time

BlazarService/src/main/java/com/hubspot/blazar/util/ModuleBuildLauncher.java
```diff
@@ -84,7 +87,7 @@ private BuildConfig fetchBuildpack(GitInfo gitInfo) throws IOException, NonRetry
   }
 
   private static BuildConfig mergeConfig(BuildConfig primary, BuildConfig secondary) {
```
tpetr: if we have this `mergeConfig()` method, then all the `BuildConfig` fields should be `Optional<>` to make merging easier

BlazarBase/src/main/java/com/hubspot/blazar/base/BuildCommand.java
jonathanwgoodwin: Why not? This way you can specify an `env` block or `returnCodes` without needing to specify your executable, which is likely to be more common than wanting to customize your executable.

BlazarBase/src/main/java/com/hubspot/blazar/base/BuildCommand.java
```diff
@@ -0,0 +1,61 @@
...
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.base.Objects;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Lists;
+
+public class BuildCommand {
+  private static final String DEFAULT_CMD = "/bin/bash";
+  private static final String DEFAULT_ARG = "-c";
```
jhaber: Should the -c be part of the default command? Otherwise people will need to add it as the first arg which seems weird, ie:

```yaml
args:
  - -c
  - mvn deploy
```

Also it seems weird that the command to run is called `args` but I'm not sure what a better name would be

BlazarService/src/main/java/com/hubspot/blazar/util/ModuleBuildLauncher.java
jhaber: They're all collections so I think its a lot nicer to just use an empty collection rather than wrap in an optional

BlazarBase/src/main/java/com/hubspot/blazar/base/BuildCommand.java
jonathanwgoodwin: Its not part of the default command if you override a field, I'll go make it do that if you didn't set the executable field.

jonathanwgoodwin: done here: https://github.com/HubSpot/Blazar/pull/46/files#diff-376fa4f392b4c2406906769c1424b7acR32

PR #47, opened 2016-01-05
Change how we're launching tasks because singuarity QA updated.

jonathanwgoodwin:


jonathanwgoodwin:
@tpetr @jhaber

BlazarService/src/main/java/com/hubspot/blazar/util/BuildLauncher.java
```diff
@@ -145,8 +147,19 @@ private synchronized void startBuild(BuildDefinition definition, Build queued, O
     LOG.info("Updating status of build {} to {}", launching.getId().get(), launching.getState());
     buildService.begin(launching);
     LOG.info("About to launch build {}", launching.getId().get());
-    HttpResponse response = asyncHttpClient.execute(buildRequest(definition, launching)).get();
-    LOG.info("Launch returned {}: {}", response.getStatusCode(), response.getAsString());
+    SingularityRunNowRequest runNowRequest = new SingularityRunNowRequest(
+        Optional.of(String.format("Build %d", launching.getId().get())),
+        Optional.of(false),
```
wsorenson: This should be absent()

wsorenson:
Do you not use runId?

PR #49, opened 2016-01-22
add correct blazar UI branch path in sidebar

zdhickman:
Broken link was missing `appRoot`. Instead of custom-building the link, just use the link passed down in `this.props` cc @benrlodge


benrlodge:
:+1: LGTM

PR #50, opened 2016-01-26
Show commit info for all build states

zdhickman:
Before:
screen shot 2016-01-26 at 11 01 07 am

After:
screen shot 2016-01-26 at 11 00 44 am

### DIFF
--- a/BlazarUI/app/scripts/components/repo-build/RepoBuildDetail.jsx
+++ b/BlazarUI/app/scripts/components/repo-build/RepoBuildDetail.jsx
@@ -79,22 +79,6 @@ class RepoBuildBuildDetail extends Component {
       durationPrefix: '',
       buildResult: humanizeText(build.state)
     }
-  
-    if (build.state === BuildStates.CANCELLED) {
-      return (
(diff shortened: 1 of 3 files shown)


jhaber:
yesssssssss

BlazarUI/app/scripts/components/repo-build/RepoBuildDetail.jsx
```diff
@@ -104,14 +88,23 @@ class RepoBuildBuildDetail extends Component {
   
     if (contains(FINAL_BUILD_STATES, build.state)) {
       buildDetail.endtime = timestampFormatted(build.endTimestamp)
+
+      if (build.state === BuildStates.CANCELLED) {
```
benrlodge: looks good. we could probably simplify this just a little more by assigning the `after` or `in` as a variable, and then only need to define `buildDetail.duration` once.

eg something like
```
const conjunction = build.state === BuildStates.CANCELLED ? 'after' : 'in';

buildDetail.duration = (
  <span>{conjunction} {build.duration}</span>
)
```

PR #52, opened 2016-01-26
Reveal org dropdown

zdhickman:
Increasing the z-index of the org dropdown has no effect, but lowering the z-indices of the build header and the build log solve the problem.


benrlodge:
looks good. yea it was probably nested too deep so wouldn't have an effect

PR #53, opened 2016-01-26
Blink on previous build statuses

zdhickman:
Might be missing some intended behavior for first time builds - will need to load in another test project.

@benrlodge I left the css extra verbose in an attempt to follow the CSS naming pattern, but it seems like there is a better way to do it.

### DIFF
--- a/BlazarUI/app/scripts/components/shared/BuildingIcon.jsx
+++ b/BlazarUI/app/scripts/components/shared/BuildingIcon.jsx
@@ -3,11 +3,13 @@ import React, {Component, PropTypes} from 'react';
 class BuildingIcon extends Component {
 
   getClassNames() {
-    if (this.props.result === "IN_PROGRESS" && this.props.prevBuildState != '') {
-      return `building-icon sidebar__active-building-icon building-icon--${this.props.result}-laststatus-${this.props.prevBuildState} ${this.props.classNames} building-icon--${this.props.size}`;
+    let prevBuildStateModifier = ``;
+
+    if (this.props.result === "IN_PROGRESS" && this.props.prevBuildState) {
+      prevBuildStateModifier = `-laststatus-${this.props.prevBuildState}`;
     }
 
-    return `building-icon sidebar__active-building-icon building-icon--${this.props.result} ${this.props.classNames} building-icon--${this.props.size}`;
+    return `building-icon sidebar__active-building-icon building-icon--${this.props.result}${prevBuildStateModifier} ${this.props.classNames} building-icon--${this.props.size}`;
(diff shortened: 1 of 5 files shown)


benrlodge:
the naming may seem verbose but I think it's actually nice and descriptive of the expected behavior, so I'd say it's good. My only tweak would be to use the [classnames](https://github.com/JedWatson/classnames) module. If you search the repo you'll find a bunch of examples.

BlazarUI/app/scripts/components/shared/BuildingIcon.jsx
```diff
@@ -3,7 +3,13 @@ import React, {Component, PropTypes} from 'react';
...
 
   getClassNames() {
-    return `building-icon sidebar__active-building-icon building-icon--${this.props.result} ${this.props.classNames} building-icon--${this.props.size}`;
+    let prevBuildStateModifier = ``;
+
+    if (this.props.result === "IN_PROGRESS" && this.props.prevBuildState) {
+      prevBuildStateModifier = `-laststatus-${this.props.prevBuildState}`;
+    }
+
+    return `building-icon sidebar__active-building-icon building-icon--${this.props.result}${prevBuildStateModifier} ${this.props.classNames} building-icon--${this.props.size}`;
```
benrlodge: using `classnames` will help make this a little easier to read.

BlazarUI/app/scripts/components/shared/BuildingIcon.jsx
```diff
@@ -3,7 +3,13 @@ import React, {Component, PropTypes} from 'react';
 class BuildingIcon extends Component {
 
   getClassNames() {
-    return `building-icon sidebar__active-building-icon building-icon--${this.props.result} ${this.props.classNames} building-icon--${this.props.size}`;
+    let prevBuildStateModifier = ``;
+
+    if (this.props.result === "IN_PROGRESS" && this.props.prevBuildState) {
```
benrlodge: i'd use the [build states](https://github.com/HubSpot/Blazar/blob/ui_v2/BlazarUI/app/scripts/constants/BuildStates.js) constants here. 

eg
`if (this.props.result === BuildStates.IN_PROGRESS && ...`

zdhickman: > Though this will look a little different once using `classnames`.

What did you have in mind? I've updated the function to use `classnames`, but the structure of the function is still similar

benrlodge: yea that looks good

PR #55, opened 2016-01-28
Docker update

jonathanwgoodwin:
Docker buildpack that we use now behaves the same way on all branches because the published image == the blazar-coordinates


jhaber:
Sweet

PR #56, opened 2016-01-28
remove faulty assumptions on buildState

zdhickman:
In the case where a build is being built for the first time, the build is in progress and there is no last build. This clears up some assumptions made on the value of `buildType`, and also no longer passes `buildType` to the `SidebarItem` component since it's not used.

cc @jonathanwgoodwin


jonathanwgoodwin:
:ship:

jhaber:
dank

PR #57, opened 2016-01-29
Star dashboard

zdhickman:
Star dashboards are so in:

screen shot 2016-01-29 at 11 28 52 am

Missing a loader. Going to throw that in, but let me know if anything else essential is missing @jhaber 

cc @benrlodge

### DIFF
--- a/BlazarUI/app/scripts/components/dashboard/StarredModulesTableRow.jsx
+++ b/BlazarUI/app/scripts/components/dashboard/StarredModulesTableRow.jsx
@@ -14,9 +14,7 @@ class StarredModulesTableRow extends Component {
 
     const item = this.props.item;
 
-    console.log(this.props.item.toString());
-
-    if (false) {
+    if (item.get('lastBuild') === undefined) {
       return (
         <tr>
           <td />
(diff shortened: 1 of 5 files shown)


BlazarUI/app/scripts/components/dashboard/StarredModulesTableRow.jsx
```diff
@@ -29,41 +39,43 @@ class StarredModulesTableRow extends Component {
...
       const gitInfo = {
-        host: latestBuildGitInfo.host,
-        organization: latestBuildGitInfo.organization,
-        repository: latestBuildGitInfo.repository
+        host: latestBuildGitInfo.get('host'),
+        organization: latestBuildGitInfo.get('organization'),
+        repository: repository
       }
 
+      latestBuild.sha = latestBuild.get('sha');
```
zdhickman: So, are all of these `get`s a necessary consequence of using `ImmutableMap`s? It worked well, save for this line where I need to force `sha` as a field in `latestBuild`, since the `Sha` component doesn't expect an `ImmutableMap` (and other components that instantiate a `Sha` do not pass an `ImmutableMap`)

benrlodge: yea, I introduced immutable after starting the project, so there's a lot of stuff that consumes regular javascript objects.  if you want a plain object you can just use `latestBuildGitInfo.toJS()`. Or you can take the time to covert over things like the `Sha` component to consume immutable objects. that would be worth doing over time.

benrlodge:
Looks good

zdhickman:
Coming back to implementing the loader later

PR #58, opened 2016-01-29
Permanent scrollbar to hopefully fix the scrolling issue

zdhickman:
screen shot 2016-01-29 at 2 53 10 pm

(if this doesn't fix the scrolling issue, then we probably have bigger problems)

cc @jhaber @benrlodge

### DIFF
--- a/BlazarUI/app/stylus/components/sidebar.styl
+++ b/BlazarUI/app/stylus/components/sidebar.styl
@@ -62,7 +62,7 @@
     width .35rem
 
   &::-webkit-scrollbar-track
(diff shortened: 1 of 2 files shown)


benrlodge:
:nice: It looks good in the image, but for some reason the scroller portion seems kind of dark and in your face when actually on the page. maybe we should lighten it up a touch? ¯\_(ツ)_/¯

zdhickman:
Lightened up:

screen shot 2016-01-29 at 4 29 30 pm

jhaber:
:hamburger:

PR #67, opened 2016-02-03
Remove zero clipboard

zdhickman:
Removing for now as opposed to fixing, since the copy buttons are probably not used much in the few places they were implemented. Could bring them back after open beta if people miss them cc @jhaber


benrlodge:
:-1:  :cry:

benrlodge:
This should be super easy to swap out with https://github.com/nkbt/react-copy-to-clipboard

jhaber:
I've never used the copy button and can't think of a time when I would

PR #72, opened 2016-02-11
Hubspot static discovery

jonathanwgoodwin:
@jhaber @also

### DIFF
--- a/BlazarService/src/main/java/com/hubspot/blazar/resources/BranchResource.java
+++ b/BlazarService/src/main/java/com/hubspot/blazar/resources/BranchResource.java
@@ -1,5 +1,16 @@
 package com.hubspot.blazar.resources;
 
+import java.io.IOException;
+import java.util.Set;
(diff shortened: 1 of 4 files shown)


BlazarService/src/main/java/com/hubspot/blazar/resources/BranchResource.java
```diff
@@ -80,4 +81,13 @@ public DiscoveryResult discoverModules(@PathParam("id") int branchId) throws IOE
...
     return result;
   }
+
+  @GET
+  @Path("{id}/discover")
+  @PropertyFiltering
+  public DiscoveryResult sideEffectFreeModuleDiscovery(@PathParam("id") int branchId) throws IOException {
+    GitInfo gitInfo = get(branchId).get();
+    return moduleDiscovery.discover(gitInfo);
+ }
```
jhaber: :+1:

PR #73, opened 2016-02-11
refresh breadcrumb paths

zdhickman:
Constructor is only called once, render is called every time we update the props - so the paths are only set at the initial page load if the breadcrumbs don't disappear while navigating the app. This fixes that issue


jhaber:
:star2:

benrlodge:
hmmmm, this works, but it's kind of a funny setup. I would move the paths object out of the `breadcrumb` component and up to the `breadcrumbs` component, generate the link there, and then just pass it down as a prop.

PR #76, opened 2016-02-12
bind this so that we can scroll up on the build log page

zdhickman:
The usual culprit cc @jhaber 

(same issue for scrolling down)


jhaber:
:art:

PR #78, opened 2016-02-12
fix table alignment to match the other tables

zdhickman:
thanks @benrlodge !


benrlodge:
i imagine you could remove the align prop all together.

zdhickman:
Fair enough, I guess I was leaning towards being explicit - but implicit left alignments are used elsewhere when `UIGridItem`s are created, so that works for me!

PR #79, opened 2016-02-12
Prevent sections w/o commands

jonathanwgoodwin:
@jhaber @gchomatas


jhaber:
`Optional.of` throws an exception on null so the preconditions check should be extraneous

jhaber:
Fixed by 4103278066c6e3011504796bdea2f419971bfbc6

PR #88, opened 2016-02-16
Support building individual modules in a branch

gchomatas:
/cc @jhaber

### DIFF
--- a/BlazarBase/src/main/java/com/hubspot/blazar/base/RepositoryBuild.java
+++ b/BlazarBase/src/main/java/com/hubspot/blazar/base/RepositoryBuild.java
@@ -1,11 +1,11 @@
 package com.hubspot.blazar.base;
 
+import java.util.Objects;
+
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.common.base.Optional;
 import com.hubspot.rosetta.annotations.StoredAsJson;
-
-import java.util.Objects;
 
 public class RepositoryBuild {
   public enum State {
@@ -35,6 +35,8 @@ public class RepositoryBuild {
   private final Optional<CommitInfo> commitInfo;
   @StoredAsJson
   private final Optional<DependencyGraph> dependencyGraph;
+  @StoredAsJson
+  private final Optional<BuildOptions> buildOptions;
 
   @JsonCreator
   public RepositoryBuild(@JsonProperty("id") Optional<Long> id,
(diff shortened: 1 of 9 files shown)


BlazarService/src/main/java/com/hubspot/blazar/listener/LaunchingRepositoryBuildVisitor.java
```diff
@@ -73,7 +75,16 @@ protected void visitLaunching(RepositoryBuild build) throws Exception {
...
     final Set<Module> toBuild = new HashSet<>();
     if (build.getBuildTrigger().getType() == Type.MANUAL) {
-      toBuild.addAll(modules);
+      if (build.getBuildTrigger().getModuleIds().isEmpty()) {
+        toBuild.addAll(modules);
+      } else {
+        final Set<Integer> moduleIds = build.getBuildTrigger().getModuleIds();
+        for (Module module : modules) {
+          if (moduleIds.contains(module.getId().or(-1))) {
+            toBuild.add(module);
```
jhaber: This won't build downstream modules of the selected ones (ie, I select `OrionData` it won't build `OrionService` as well). Maybe whether to include downstreams should be an additional option in the modal window the user gets? At that point maybe we should move these fields from the `BuildTrigger` object into a new `BuildOptions` object because @also wanted some options as well (enable SSH access, clear the cache directories, etc.) which could get added to this modal and to the `BuildOptions` object.

This is a more involved change (new DB column, new field on the RepositoryBuild) but I think it might be better in the long run

gchomatas:
np I will do the changes

BlazarBase/src/main/java/com/hubspot/blazar/base/RepositoryBuild.java
```diff
@@ -108,29 +112,37 @@ public BuildTrigger getBuildTrigger() {
...
-    return new RepositoryBuild(id, branchId, buildNumber, state, buildTrigger, startTimestamp, endTimestamp, sha, Optional.of(commitInfo), dependencyGraph);
+    return new RepositoryBuild(id, branchId, buildNumber, state, buildTrigger, startTimestamp, endTimestamp, sha, Optional.of(commitInfo), dependencyGraph, buildOptions);
   }
 
   public RepositoryBuild withDependencyGraph(DependencyGraph dependencyGraph) {
-    return new RepositoryBuild(id, branchId, buildNumber, state, buildTrigger, startTimestamp, endTimestamp, sha, commitInfo, Optional.of(dependencyGraph));
+    return new RepositoryBuild(id, branchId, buildNumber, state, buildTrigger, startTimestamp, endTimestamp, sha, commitInfo, Optional.of(dependencyGraph), buildOptions);
+  }
+
+  public RepositoryBuild withBuildOptions(BuildOptions buildOptions) {
```
jhaber: Is this method ever used?

BlazarService/src/main/java/com/hubspot/blazar/listener/LaunchingRepositoryBuildVisitor.java
```diff
@@ -68,41 +70,55 @@ protected void visitLaunching(RepositoryBuild build) throws Exception {
...
+      if (!build.getBuildOptions().isPresent() || build.getBuildOptions().get().getModuleIds().isEmpty()) {
+        toBuild.addAll(allModules);
+      } else {
+        final Set<Integer> requestedModuleIds = build.getBuildOptions().get().getModuleIds();
+        for (Module module : allModules) {
+          if (requestedModuleIds.contains(module.getId().get())) {
+            toBuild.add(module);
+          }
+        }
+        addDownstreamModules(build, allModules, toBuild);
```
jhaber: We should check if this build option is enabled before adding downstreams right?

gchomatas: Sure I already added an option for that :)

PR #89, opened 2016-02-17
Debug query param for singularity task link

zdhickman:
Now, default behavior on a repo build page is to _not_ show the singularity task. If you add `?debug` to the end of the url, then the task link will appear. cc @jonathanwgoodwin 

(this is very much temporary)

### DIFF
--- a/BlazarUI/app/scripts/components/repo-build/RepoBuildModulesTableRow.jsx
+++ b/BlazarUI/app/scripts/components/repo-build/RepoBuildModulesTableRow.jsx
@@ -21,7 +21,15 @@ class RepoBuildModulesTableRow extends Component {
     );    
   }
 
+  isDebugMode() {
+    return window.location.href.indexOf('?debug') > -1;
+  }
+
   renderSingularityLink() {
+    if (!this.isDebugMode()) {
+      return null;
+    }
+
     const {taskId} = this.props.data;
 
     if (!taskId) {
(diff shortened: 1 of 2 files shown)


jonathanwgoodwin:
:fire: :droplet: :boat:

BlazarUI/app/scripts/components/repo-build/RepoBuildModulesTableRow.jsx
```diff
@@ -21,7 +21,15 @@ class RepoBuildModulesTableRow extends Component {
     );    
   }
 
+  isDebugMode() {
+    return window.location.href.indexOf('?debug') > -1;
```
benrlodge: better to handle this in the react-router and pass down as props

benrlodge: https://github.com/reactjs/react-router/blob/latest/examples/query-params/app.js#L8-L9

zdhickman: I know :disappointed:  I started the nice solution through the router and changed my mind, since this is a temporary change. Once we look at the auth'd user cookie, we'll probably compare to a list of users to determine if we should show the Singularity link. But thanks for pointing it out

benrlodge:
actually... on that note anyways - I realized the example if for v1 but blazar is still using v0.13 - probably worth upgrading at some point.

PR #92, opened 2016-02-18
Module modal

zdhickman:
This PR introduces the feature for selecting specific modules to build, when clicking `Build Now` on a branch build history page. Additionally, there is a toggle for whether or not to include building downstream modules:

screen shot 2016-02-22 at 1 50 50 pm

TODO:
- [x] Code clean-up
- [x] Better style/UX in the modal
- [x] Loader in module modal (extra)
- [x] Improved styling in modal

cc @jhaber @jonathanwgoodwin @gchomatas


BlazarUI/app/scripts/components/branch/BranchContainer.jsx
```diff
@@ -105,10 +122,16 @@ class BranchContainer extends Component {
             </UIGridItem>
             <UIGridItem size={2} align='RIGHT'>
               <BuildButton 
-                triggerBuild={this.triggerBuild} 
+                triggerBuild={this.openModuleModal.bind(this)}
```
benrlodge: you can simplify/keep all these bindings clean and packaged up by using `_.bindAll` in the constructor - you can see examples of this across the app.

BlazarUI/app/scripts/components/shared/ModuleModal.jsx
```diff
@@ -0,0 +1,83 @@
...
+    return (
+      <Checkbox
+        label='Build Downstream Modules'
+        name='downstream-checkbox' />
+    );
+  }
+
+  renderModuleList() {
+    return (
+      <div className='module-modal--container'>
```
benrlodge: if you're getting at using the BEM syntax (here and below), the `--` should represent a modifier, not element, so it looks like you'd want `module-modal` as the block, `module-module__container` as the element. though in this case I would just use something like `module-modal` (dropping `container`), and with the elements below, `module-modal__modules` `module-modal__downstream`, etc. Keep `--` as modifiers, eg if you were to have `modal-modal__title` as BE, a modifier would look like `modal-modal__title--big` `modal-modal__title--small` for arbitrary example.

BlazarUI/app/scripts/components/shared/ModuleModal.jsx
```diff
@@ -0,0 +1,83 @@
...
+        <div className='module-modal--modules'>
+          <span className='module-modal--title'>Pick Modules</span>
+          {this.props.modules.map(this.renderModule)}
+        </div>
+        <div className='module-modal--downstream'>
+          <span className='module-modal--title'>Toggle Option</span><br />
```
benrlodge: not that `<br>` tags are wrong, I'd stick with using `<p>` tags if you want to create a block, and CSS for spacing tweaks, for the most part.

BlazarUI/app/scripts/components/shared/ModuleModal.jsx
```diff
@@ -0,0 +1,83 @@
...
+          {this.renderDownstreamToggle()}
+        </div>
+      </div>
+    );
+  }
+
+  getModuleIdsAndBuild() {
+    // TODO: more specific/no $
```
benrlodge: !

BlazarUI/app/stylus/components/module-modal.styl
```diff
@@ -0,0 +1,12 @@
+.module-modal--container
+  height 200px
+
+.module-modal--modules
+  float left
```
benrlodge: i would just keep this simple and use the bootstrap grid, no need for custom css

BlazarUI/app/stylus/components/module-modal.styl
```diff
@@ -0,0 +1,12 @@
...
+
+.module-modal--downstream
+  float right
+
+.module-modal--title
+  font-size 1.5em
```
benrlodge: same as ^, i would either use `h` tags, or if you want to have custom modal styles, I would make this more general for all modals, eg `modal-title`, so you're not writing custom css for every different modal that exists.

BlazarUI/app/stylus/main.styl
```diff
@@ -30,6 +30,7 @@
 @import 'components/icon'
 @import 'components/json-modal'
 @import 'components/loader'
+@import 'components/module-modal'
```
benrlodge: ...thinking about it some more yea i would definitely remove these styles all together and just stick w/bootstrap to keep it simple

BlazarUI/app/scripts/components/shared/ModuleModal.jsx
zdhickman: Thanks for the BEM reminder

 to this and your other comments re: css and styling, I intend to overhaul the entire modal format anyway. Mostly likely all of it will be bootstrap anyway

PR #103, opened 2016-02-23
running timer on branch build history page

zdhickman:
Any idea on how to do this better? Running timer for in progress builds. There's not much of a state to update higher up in the hierarchy, since the only thing changing is the result of `moment()`. /cc @benrlodge

### DIFF
--- a/BlazarUI/app/scripts/components/branch/BranchBuildHistoryTableRow.jsx
+++ b/BlazarUI/app/scripts/components/branch/BranchBuildHistoryTableRow.jsx
@@ -10,8 +10,27 @@ import {tableRowBuildState, timestampFormatted, humanizeText, buildResultIcon, g
 import Icon from '../shared/Icon.jsx';
 import Sha from '../shared/Sha.jsx';
 
+let initialState = {
+  moment: moment()
+}
 
 class BranchBuildHistoryTableRow extends Component {
+
+  constructor(props) {
+    super(props);
+
+    this.state = initialState;
+  }
+
+  componentDidMount() {
+    this.interval = setInterval(this.updateMoment.bind(this), 1000);
+  }
+
+  updateMoment() {
+    this.setState({
+      moment: moment()
+    })
+  }
 
   onTableClick(e) {
     if (e.target.className === 'sha-link') {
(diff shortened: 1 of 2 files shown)


BlazarUI/app/scripts/components/branch/BranchBuildHistoryTableRow.jsx
```diff
@@ -68,6 +74,10 @@ class BranchBuildHistoryTableRow extends Component {
   render() {
     const {data, params} = this.props;
 
+    setTimeout(() => {
+      this.forceUpdate();
```
benrlodge: we should avoid this

benrlodge:
can you explain what you're trying to achieve in this PR?

benrlodge:
Isn't the build history container polling for updates? I believe it was, or should have been. So if it was, the row component should be re-rendering with the latest `renderDuration` calculation. It seems like I may be missing something here though?

BlazarUI/app/scripts/components/branch/BranchBuildHistoryTableRow.jsx
zdhickman: I intend to

zdhickman:
This is an example page that the change would affect: https://private.hubteam.com/blazar/builds/git.hubteam.com/HubSpot/Ads/master

We want to add a running counter for in progress builds, like I've already added here: https://github.com/HubSpot/Blazar/pull/101

The build history container polls for updates every 5 seconds. This means there is a re-render for a row every 5 seconds. I'd rather have the re-render occur every second in the name of good UX.

I don't want to use force update. The equivalent, possibly "better" React solution is to have a `setTimeout()`-calling function up in the container that updates the state each second, but that also seems hacky considering the only state change is the current result of `moment()`. So, now I'm wondering how to get the functionality covered by this PR in its current state, without using `forceUpdate()`.

zdhickman:
```
updateMoment() {
  this.setState({
    currentTimestamp: moment()
  });

  setTimeout(() => {
    this.updateMoment();
  }, 1000);
}
```

zdhickman:
I updated to a slightly better use of `setTimeout()` but I'm still falsely updating state

PR #104, opened 2016-02-24
A better org page

zdhickman:
Before:

screen shot 2016-02-24 at 11 25 46 am

After:

screen shot 2016-02-24 at 11 25 55 am


jhaber:
so epic :fire:

PR #106, opened 2016-02-24
Slack

jonathanwgoodwin:
@jhaber @gchomatas


BlazarData/src/main/java/com/hubspot/blazar/data/dao/SlackConfigurationDao.java
```diff
@@ -0,0 +1,41 @@
...
+import org.skife.jdbi.v2.sqlobject.SqlUpdate;
+import org.skife.jdbi.v2.sqlobject.customizers.SingleValueResult;
+
+import com.google.common.base.Optional;
+import com.hubspot.blazar.base.slack.SlackConfiguration;
+import com.hubspot.rosetta.jdbi.BindWithRosetta;
+
+public interface SlackConfigurationDao {
+
+  @SqlQuery("SELECT * FROM slack_configs where active = 1")
```
zdhickman: `WHERE`

BlazarService/src/main/java/com/hubspot/blazar/listener/SlackNotificationVisitor.java
```diff
@@ -0,0 +1,188 @@
...
+    // OnFailure
+    if (slackConfiguration.getOnFail() && FAILED_REPO_STATES.contains(state)) {
+      return true;
+    }
+    return false;
+  }
+
+  private void sendSlackMessage(SlackConfiguration slackConfiguration, RepositoryBuild build, GitInfo gitInfo) {
+    String fallback = String.format("Module Build %s-%s-%s finished with state %s", gitInfo.getRepository(), gitInfo.getRepository(), gitInfo.getBranch(), build.getState().toString());
+    Optional<String> color = Optional.absent();
```
zdhickman: you have `ABSENT_STRING` defined as a constant and used below - might as well use it here too?

BlazarService/src/main/java/com/hubspot/blazar/listener/SlackNotificationVisitor.java
```diff
@@ -0,0 +1,199 @@
...
+    this.slackClient = slackClient;
+    this.repositoryBuildService = repositoryBuildService;
+  }
+
+
+  @Override
+  public void visit(RepositoryBuild build) throws Exception {
+    Set<SlackConfiguration> configurationSet = slackConfigurationService.getAllWithBranchId(build.getBranchId());
+    for (SlackConfiguration slackConfiguration : configurationSet) {
+      Optional<RepositoryBuild> previous = repositoryBuildService.getPreviousBuild(build);
```
zdhickman: Does the result of `repositoryBuildService.getPreviousBuild(build)` change based on the current `configurationSet` you're iterating on? If not, can you run the `previous.isPresent()` step outside of this `for` loop so that you can exit quickly?

jonathanwgoodwin:
rebased off of master to get changes to ModuleBuildDao & Service

BlazarService/src/main/java/com/hubspot/blazar/resources/FeedbackResource.java
```diff
@@ -9,35 +11,44 @@
...
+import com.hubspot.blazar.externalservice.slack.SlackMessage;
 import com.hubspot.blazar.base.feedback.Feedback;
-import com.hubspot.blazar.config.BlazarConfiguration;
-import com.hubspot.blazar.config.SlackConfiguration;
-import in.ashwanthkumar.slack.webhook.Slack;
-import in.ashwanthkumar.slack.webhook.SlackAttachment;
+import com.hubspot.blazar.config.BlazarSlackConfiguration;
+import com.hubspot.blazar.integration.slack.SlackClient;
 
 @Path("/feedback")
```
gchomatas: I assume this is meant to capture some user feedback. Maybe rename the class to UserFeedback and the end-point /user-feedback

BlazarService/src/main/java/com/hubspot/blazar/resources/InstantMessageResource.java
```diff
@@ -0,0 +1,88 @@
...
+
+  @PUT
+  @Path("/configurations")
+  public Optional<InstantMessageConfiguration> insert(InstantMessageConfiguration instantMessageConfiguration) {
+    long id = instantMessageConfigurationService.insert(instantMessageConfiguration);
+    return instantMessageConfigurationService.get(id);
+  }
+
+  @POST
+  @Path("/configurations")
```
gchomatas: Conventionally you POST to insert a new resource object and you PUT (is idempotent) to an ID to update an existing resource with that Id. So when you PUT you better go for: 
PUT /configurations/id (more clean that you update an existing resource)
You can also PUT /configurations as you do above but you need to check if the posted object has an id already and throw an exception if it hasn't, I would prefer the former approach.

BlazarService/src/main/java/com/hubspot/blazar/resources/InstantMessageResource.java
```diff
@@ -0,0 +1,88 @@
...
+  public Optional<InstantMessageConfiguration> update(InstantMessageConfiguration instantMessageConfiguration) {
+    instantMessageConfigurationService.update(instantMessageConfiguration);
+    return instantMessageConfigurationService.get(instantMessageConfiguration.getId().get());
+  }
+
+  @DELETE
+  @Path("/configurations")
```
gchomatas: What about having the id in the path? 
DELETE /configurations/{id}

BlazarService/src/main/java/com/hubspot/blazar/resources/FeedbackResource.java
jonathanwgoodwin: this requires an associated UI change

gchomatas:
Other than the above comments LGTM

BlazarService/src/main/java/com/hubspot/blazar/resources/InstantMessageResource.java
jonathanwgoodwin: :+1:

PR #109, opened 2016-02-29
Upgrade react

zdhickman:
Hit some functionality limitations on the old packages, so went through two upgrades from React 0.x -> 1.x -> 2.x, as well as other peer dependency upgrades. Then, cleaning up all the console errors and warnings from the upgrade


also:
React 1.x?

zdhickman:
> React 1.x?

oop, `react-router` 1.x

PR #110, opened 2016-02-29
Metrics

jonathanwgoodwin:


BlazarService/src/main/java/com/hubspot/blazar/listener/MetricBuildVisitor.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    for (RepositoryBuild.State state : RepositoryBuild.State.values()) {
+      metricRegistry.register(RepositoryBuild.class.getName() + "." + state.toString().toLowerCase(), new Meter());
+    }
+  }
+
+  public void visit(ModuleBuild moduleBuild) {
+    SortedMap<String, Meter> meters = metricRegistry.getMeters();
+    meters.get(moduleBuild.getClass().getName() + "." + moduleBuild.getState().toString().toLowerCase()).mark();
```
jhaber: I prefer `name()` over `toString()` for this in case someone overrides `toString()`. Also you don't need to pre-register the meters, you can just do something like `metricRegistry.meter(metricName).mark()` inside this method and remove the pre-registration form the constructor

jonathanwgoodwin: Oh we don't need to register?! :dancer:

jhaber: You can still pre-register if you want the meters to always be returned but in either case you can do something like `metricRegistry.meter(metricName).mark()` inside this method

PR #111, opened 2016-02-29
more metrics

jonathanwgoodwin:


BlazarService/src/main/java/com/hubspot/blazar/util/SingularityBuildWatcher.java
```diff
@@ -97,6 +103,12 @@ public void run() {
                 String taskId = task.get().getTaskId().getId();
                 LOG.info("Updating build {} to FAILED because taskId {} is done", build.getId().get(), taskId);
                 moduleBuildService.update(build.withState(State.FAILED).withTaskId(taskId).withEndTimestamp(task.get().getUpdatedAt()));
+                if (taskState.get().isSuccess()) {
+                  metricRegistry.meter(getClass().getName() + ".succeeded").mark();
+                } else {
+                  metricRegistry.meter(getClass().getName() + ".failed").mark();
+                }
```
jhaber: This should already be handled by the visitor

jhaber: Oh is this supposed to indicate an abnormal success/failure?

jonathanwgoodwin: Success here is a weird case, it implies that the executor didn't report back, but exited 0, I think its good to track that just in case. Failure of course is more expected.

PR #117, opened 2016-03-03
(WIP) New sidebar

zdhickman:
Old sidebar: 
screen shot 2016-03-03 at 1 23 27 pm

New sidebar:
screen shot 2016-03-03 at 1 23 09 pm

Changes:
- Upped font size from 11px -> 16px
- Combined branches of a common repo into a single sidebar item
- Only display non-success build status icons next to branches
- Inline expanding of sidebar items with >3 branches:

screen shot 2016-03-03 at 1 26 10 pm

screen shot 2016-03-03 at 1 25 50 pm

TODO:
- [x] Copy on `show/hide` text
- [x] Code clean-up
- [x] Correctly colored links
- [x] Touch up styling

### DIFF
--- a/BlazarUI/app/scripts/components/shared/LazyRender.jsx
+++ b/BlazarUI/app/scripts/components/shared/LazyRender.jsx
@@ -6,12 +6,14 @@ let LazyRender = React.createClass({
     maxHeight: React.PropTypes.number.isRequired,
     childHeight: React.PropTypes.number,
     className: React.PropTypes.string,
-    itemPadding: React.PropTypes.number
+    itemPadding: React.PropTypes.number,
+    ignoreChildHeight: React.PropTypes.bool
   },
 
   getDefaultProps: function() {
     return {
-      itemPadding: 7
+      itemPadding: 7,
+      ignoreChildHeight: false
     };
   },
 
@@ -44,6 +46,14 @@ let LazyRender = React.createClass({
   },
 
   getHeight: function(numChildren, childHeight, maxHeight) {
+    if (this.props.ignoreChildHeight) {
+      return maxHeight;
+    }
+
+    let fullHeight = numChildren * childHeight;
+    if (fullHeight < maxHeight) {
+      return fullHeight;
+    }
     return maxHeight;
   },
 
--- a/BlazarUI/app/scripts/components/shared/LazyRender.jsx
+++ b/BlazarUI/app/scripts/components/shared/LazyRender.jsx
@@ -44,10 +44,6 @@ let LazyRender = React.createClass({
   },
 
   getHeight: function(numChildren, childHeight, maxHeight) {
-    let fullHeight = numChildren * childHeight;
-    if (fullHeight < maxHeight) {
-      return fullHeight;
-    }
     return maxHeight;
   },
 
--- a/BlazarUI/app/scripts/components/sidebar/SidebarItem.jsx
+++ b/BlazarUI/app/scripts/components/sidebar/SidebarItem.jsx
@@ -63,7 +63,7 @@ class SidebarItem extends Component {
       <div className='sidebar-item__repo-link'>
         <Icon type='octicon' name='repo' classNames='repo-octicon'/>{ '   ' }
           <span className='sidebar-item__module-repo-name'>
-            {truncate(repository, 30, true)}
+            {truncate(repository, 25, true)}
           </span>
       </div>
     );
--- a/BlazarUI/app/scripts/components/sidebar/SidebarItem.jsx
+++ b/BlazarUI/app/scripts/components/sidebar/SidebarItem.jsx
@@ -51,7 +51,7 @@ class SidebarItem extends Component {
 
     return (
       <div onClick={this.toggleExpand.bind(this)} className='sidebar-item__and-more'>
-        show {numberToHide} less
+        show fewer
       </div>
     );
   }
@@ -63,7 +63,7 @@ class SidebarItem extends Component {
       <div className='sidebar-item__repo-link'>
         <Icon type='octicon' name='repo' classNames='repo-octicon'/>{ '   ' }
           <span className='sidebar-item__module-repo-name'>
-            {truncate(repository, 25, true)}
+            {truncate(repository, 20, true)}
           </span>
       </div>
     );
--- a/BlazarUI/app/scripts/components/sidebar/SidebarItem.jsx
+++ b/BlazarUI/app/scripts/components/sidebar/SidebarItem.jsx
@@ -1,4 +1,5 @@
 import React, {Component, PropTypes} from 'react';
+import {Link} from 'react-router'
 import classnames from 'classnames';
 import {has, contains} from 'underscore';
 import {truncate} from '../Helpers.js';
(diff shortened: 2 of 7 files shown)


BlazarUI/app/scripts/components/shared/LazyRender.jsx
```diff
@@ -44,10 +44,6 @@ let LazyRender = React.createClass({
   },
 
   getHeight: function(numChildren, childHeight, maxHeight) {
-    let fullHeight = numChildren * childHeight;
-    if (fullHeight < maxHeight) {
-      return fullHeight;
-    }
     return maxHeight;
   },
```
benrlodge: I would pass this in as an optional prop, otherwise it limits the use of this component elsewhere.

BlazarUI/app/scripts/components/sidebar/SidebarItem.jsx
```diff
@@ -1,92 +1,161 @@
...
+          show {numberRemaining} more
+        </div>
+      );
+    }
+
+    const numberToHide = builds.length - 3;
+
+    return (
+      <div onClick={this.toggleExpand.bind(this)} className='sidebar-item__and-more'>
+        show {numberToHide} less
```
benrlodge: Because the two last returns only differ by the text value, I would consider writing one condition for the text so then you only need to return one component.

eg something like
```
const numberMessage = !this.state.expanded ? `${numberRemaing} more` : `${builds.length - 3} less`;

<div onClick={this.toggleExpand.bind(this)} className='sidebar-item__and-more'>
         show {numberMessage}

...snip...

```

BlazarUI/app/scripts/components/sidebar/SidebarItem.jsx
```diff
@@ -1,92 +1,161 @@
...
+          </span>
       </div>
     );
   }
 
-  renderBranchText() {
-    const {gitInfo} = this.props;
+  renderBranchText(build) {
+    const gitInfo = build.gitInfo;
+    const lastBuild = build.lastBuild;
```
benrlodge: you can destructure these -

`const {gitInfo, lastBuild} = build;`

BlazarUI/app/scripts/components/sidebar/SidebarItem.jsx
```diff
@@ -1,92 +1,161 @@
...
+    return (
+      <div className='sidebar-item__building-icon-link'>
+        {buildResultIcon(buildState)}
+      </div>
+    );
+  }
+
+  renderBranchRow(build) {
+    const gitInfo = build.gitInfo;
+    const lastBuild = build.lastBuild;
```
benrlodge: destructure

BlazarUI/app/scripts/components/sidebar/SidebarItem.jsx
```diff
@@ -1,92 +1,161 @@
...
+          {this.renderExpandText(numberRemaining)}
+        </div>
+      );
+    } 
+
+    else {
+      return (
+        <div>
+          {realBuilds.map((build) => {return this.renderBranchRow(build);})}
+          {this.renderExpandText(0)}
```
benrlodge: not a huge deal but you can use an ES6 default function parameter in `renderExpandText` so you don't need to pass 0 here.

BlazarUI/app/scripts/components/sidebar/SidebarRepoList.jsx
```diff
@@ -21,16 +21,29 @@ class SidebarRepoList extends Component {
       );
     }
 
-    const buildsList = sortBuildsByRepoAndBranch(filterInactiveBuilds(filteredBuilds)).map( (build) => {
-      const buildType = has(build, 'inProgressBuild') ? 'inProgressBuild' : has(build, 'lastBuild') ? 'lastBuild' : 'neverBuilt'
+
+
+    const buildsList = Object.keys(filteredBuilds).sort().map((repo) => {
```
benrlodge: this is turning into a hefty render method, I would extract some of this logic out of it.

BlazarUI/app/scripts/utils/buildsHelpers.js
```diff
@@ -49,3 +49,26 @@ export const sortBuilds = (builds, type) => {
   }
 
 };
+
+export const sidebarCombine = (builds) => {
+  let sidebarMap = {};
+
+  builds.map((build) => {
+    const repo = build.gitInfo.repository;
+    const branch = build.gitInfo.branch;
```
benrlodge: can destructure

BlazarUI/app/stylus/components/icon.styl
```diff
@@ -28,8 +28,15 @@
...
 .sidebar-item__building-icon-link > .table-icon-container > span > i
-  width 35px
+  width 38px
+  margin-top -8px
 
 .fa-stop
   font-size .55em
   line-height 2.7
+
+.sidebar-item > * > .octicon-repo
```
benrlodge: this is pretty funky. are there multiple levels of `octicon-repo` being used in the sidebar, and you're trying to bypass one? I don't think so right? Can you just do `.sidebar-item .octicon-repo` ?

benrlodge:
Looking good. I'm wondering if we may want to try to break up `SidebarItem` a little since it's starting to grow pretty big. Maybe the expanding aspect can be moved out to a separate component? Just a thought.

BlazarUI/app/scripts/components/sidebar/SidebarItem.jsx
zdhickman: For sure! Going to avoid the one liner and the cleverness of `show` separated from the rest of message in favor of the slightly more readable and verbose series of `if`s

BlazarUI/app/stylus/components/icon.styl
zdhickman: Thanks - I just forgot how CSS worked at the time :lock_with_ink_pen:

BlazarUI/app/scripts/components/sidebar/SidebarItem.jsx
benrlodge: You can still do `if`s as opposed to a ternary to generate the text, if you find that more readable. My main point is that there is unnecessary duplicate jsx and returns -

`<div onClick={this.toggleExpand.bind(this)} className='sidebar-item__and-more'>`

zdhickman:
Thanks for the comments :v: 

We can get away with fitting this new format of repo-branch build organization in the sidebar, by manipulating the format of the `builds` param passed down the sidebar components. A rewrite of these components will be necessary if we mess with the sidebar any more (which may happen in the near future). Eventually, I would restructure the components to match the format of the new `builds` object (with further subdivisions like branch row components, etc.)

BlazarUI/app/scripts/components/shared/LazyRender.jsx
zdhickman: Yeah I made this change last minute to fix a bug. What do you want passed in as an optional prop?

BlazarUI/app/scripts/components/sidebar/SidebarItem.jsx
zdhickman: Gotcha - that has been fixed

PR #119, opened 2016-03-07
(WIP) Show all commits for a build

zdhickman:
screen shot 2016-03-07 at 3 42 47 pm


BlazarUI/app/scripts/components/repo-build/Commit.jsx
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  renderContent() {
+    const {message, author, modified, added, removed, url} = this.props.commitInfo;
+    const {name} = author;
+    const filesChanged = modified.length + added.length + removed.length;
+
+    return (
+      <div className="commits-container--commit-wrapper">
+        <div className="commits-container--commit-content">
```
benrlodge: BEM nitpick throughout the file, element vs modifier
```
 <div className="commits-container">
    <div className="commits-container__commit-content">
    <span className="commits-container__commit-message">
```
etc.

and for brevity I would suggest
```
 <div className="commits-container">
    <div className="commits-container__content">
    <span className="commits-container__message">
```

BlazarUI/app/scripts/components/repo-build/RepoBuildDetail.jsx
```diff
@@ -111,16 +113,8 @@ class RepoBuildBuildDetail extends Component {
...
-          <pre className='build-detail-body__commit-desc' title={currentCommit.message}>{truncate(currentCommit.message, this.state.windowWidth * .08, true)}</pre>
-        </div>
-        <div className='build-detail-footer'>
-          Triggered by <strong>{currentCommit.author.name}</strong> on { ' ' }
-          {timestampFormatted(currentCommit.timestamp, 'llll')} 
-          <span className='build-detail__sha'> 
-            commit <Sha gitInfo={gitInfo} build={build} />
-          </span>
-        </div>
+        <CommitsContainer
```
benrlodge: I would suggest avoiding the use of `container` in the naming of this component for consistency in how `container` is used across the app. The app's current use of `container` is only used for a top level component that handle data fetching and state for its children (eg https://medium.com/@learnreact/container-components-c0e67432e005#.ifyto7j1n). That's not the case here, so I would suggest naming it something like `commits` and then renaming the current `commits` component to `commit-detail` or `commit-list` or something like that.

BlazarUI/app/scripts/components/repo-build/Commit.jsx
zdhickman: re: the first part - I was so close :(

BlazarUI/app/scripts/components/repo-build/RepoBuildDetail.jsx
zdhickman: :+1:

PR #120, opened 2016-03-08
Update deps to reference new locations.

kenbreeman:
@tpetr @jhaber 

Feels weird having these here, thoughts?


tpetr:
These strings must go away, especially if we want anyone else to be able to use this. @jonathanwgoodwin, @jhaber, or @gchomatas should be able to square this away.

jhaber:
The strings are staying, we just need to move these classes out into their own JAR that gets deployed with Blazar.

PR #123, opened 2016-03-10
Surface malformed config files in branch

zdhickman:
Bootstrap alerts in-app:

screen shot 2016-03-10 at 9 31 38 am
screen shot 2016-03-10 at 9 31 50 am

----

The modal after clicking the alert:

screen shot 2016-03-10 at 9 31 55 am

----

Please help with copy! @jhaber @jonathanwgoodwin


zdhickman:
I've duplicated the api call to grab malformed file info because the respective apis for each container both have their own ways of determining the branch and using that value later on in the api. There's room for api call optimizations as more endpoints have data useful to multiple containers

jhaber:
:hamburger:

PR #125, opened 2016-03-11
Improve tables

zdhickman:
Most tables have a link to most recent build, so that's what I set the default link to be for each row. That applies to all table rows except for the repo-build-module table, which will link to the module's build log

### DIFF
--- a/BlazarUI/app/scripts/components/dashboard/StarredBranchesTableRow.jsx
+++ b/BlazarUI/app/scripts/components/dashboard/StarredBranchesTableRow.jsx
@@ -51,23 +51,12 @@ class StarredBranchesTableRow extends Component {
       sha = <Sha gitInfo={gitInfo} build={latestBuild} />;
     }
 
-    let currentState;
-    let previousState;
-
-    if (item.get('inProgressBuild') !== undefined) {
-      currentState = item.get('inProgressBuild').get('state');
-      previousState = latestBuild.get('state');
-    }
-
-    else {
-      currentState = latestBuild.get('state');
-      previousState = '';
-    }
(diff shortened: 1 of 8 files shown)


BlazarUI/app/scripts/components/dashboard/StarredBranchesTableRow.jsx
```diff
@@ -9,6 +10,30 @@ import CommitMessage from '../shared/CommitMessage.jsx';
...
+    ]);
+  }
+
+  onTableClick(blazarBranchPath, blazarPath, e) {
+    if (e.target.className === 'sha-link') {
+      window.open(e.target.href, '_blank');
+    }
+
+    else if (e.target.className === 'repo-link') {
+      this.context.router.push(blazarBranchPath);
```
benrlodge: now that we've upgraded react-router to v2 we should replace these with browserHistory

`browserHistory.push(blazarBrachPath)`

BlazarUI/app/scripts/components/dashboard/StarredBranchesTableRow.jsx
```diff
@@ -9,6 +10,30 @@ import CommitMessage from '../shared/CommitMessage.jsx';
...
+
+  getRowClassNames(state) {
+    return classNames([
+      tableRowBuildState(state),
+      'clickable-table-row'
+    ]);
+  }
+
+  onTableClick(blazarBranchPath, blazarPath, e) {
+    if (e.target.className === 'sha-link') {
```
benrlodge: it's a little more robust to pass this className value in as another argument. In general it's best to keep class names strictly for style, and not mix them up with the logic. Helps minimize potential issues in the future. 

You can also use underscores [partial](http://underscorejs.org/#partial) in these types of cases rather than bind if the context isn't necessary.

PR #126, opened 2016-03-11
make queued build page render + cancel queued builds

zdhickman:


BlazarUI/app/scripts/components/repo-build/RepoBuildCancelButton.jsx
```diff
@@ -27,7 +28,10 @@ class CancelBuildButton extends Component {
     this.setState({
       cancelling: true,
       showModal: false
-    });    
+    });
+
+    const parentPath = window.location.pathname.slice(0, window.location.pathname.lastIndexOf('/'));
```
zdhickman: @benrlodge any idea how to accomplish this without `window.location.pathname`? I'm trying to get back to the parent component - the branch build history page

benrlodge: use [browserHistory](https://github.com/reactjs/react-router/blob/master/docs/guides/Histories.md)

```
import { browserHistory } from 'react-router'
```

then you can trigger - 

```
browserHistory.goBack();
```

benrlodge: you can also do it one better and create a custom reusable history util module to store these types of functions, and import it where you need it.

benrlodge: On second thought, this assumes the user came from the browser history page.

benrlodge: Just pass the `this.props.params` prop down to the build page (starting from the container) and you can build the route from that.

PR #128, opened 2016-03-14
Static deps & provides

jonathanwgoodwin:
@also @gchomatas @jhaber

### DIFF
--- a/BlazarService/src/main/java/com/hubspot/blazar/discovery/hsstatic/HubSpotStaticModuleDiscovery.java
+++ b/BlazarService/src/main/java/com/hubspot/blazar/discovery/hsstatic/HubSpotStaticModuleDiscovery.java
@@ -12,9 +12,12 @@ import javax.inject.Singleton;
 import org.kohsuke.github.GHRepository;
 import org.kohsuke.github.GHTree;
 import org.kohsuke.github.GHTreeEntry;
+import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


jhaber:
Do we need to include runtimeDeps or devDeps as well?

jonathanwgoodwin:
Oh yeah. I'll add post lunch.

BlazarService/src/main/java/com/hubspot/blazar/discovery/hsstatic/HubSpotStaticModuleDiscovery.java
```diff
@@ -60,21 +63,23 @@ public DiscoveryResult discover(GitInfo gitInfo) throws IOException {
     Set<DiscoveredModule> modules = new HashSet<>();
     Set<MalformedFile> malformedFiles = new HashSet<>();
     for (String config : staticConfigs) {
-      String moduleDirectory = String.valueOf(Paths.get(config).getParent());
-      String moduleName = moduleName(gitInfo, moduleDirectory);
+      String contents = gitHubHelper.contentsFor(config, repository, gitInfo);
+      StaticConfig configObject = objectMapper.readValue(contents, StaticConfig.class);
```
jhaber: We should add a malformed file if this parsing fails

jonathanwgoodwin:
@jhaber We good here?

jhaber:
cool lgtm

PR #131, opened 2016-03-16
Eagerly launch Singularity tasks for each module build

jhaber:
Executor has been updated to wait for the build to enter the launching state before continuing, but this will reduce the overhead by having the docker container pulled and launched, JVM started, etc. so once the build enters the launching state we can go right away

@jonathanwgoodwin @gchomatas


jonathanwgoodwin:
:ghost: sounds good!

PR #139, opened 2016-03-23
fix some logics

jonathanwgoodwin:


BlazarService/src/main/java/com/hubspot/blazar/listener/SlackNotificationVisitor.java
```diff
@@ -66,6 +66,9 @@ public SlackNotificationVisitor(InstantMessageConfigurationService instantMessag
   @Override
   public void visit(RepositoryBuild build) throws Exception {
     Set<InstantMessageConfiguration> configurationSet = instantMessageConfigurationService.getAllWithBranchId(build.getBranchId());
+    if (!(build.getState().isComplete())) {
+      return;
+    }
```
jhaber: Maybe do this before you fetch configs from the DB?

PR #140, opened 2016-03-23
add # because private channls dont require but public ones must have

jonathanwgoodwin:


BlazarService/src/main/java/com/hubspot/blazar/externalservice/slack/SlackApiResponse.java
```diff
@@ -38,7 +40,7 @@ public String getError() {
   public String toString() {
     return Objects.toStringHelper(this)
         .add("ok", ok)
-        .add("error", error)
+        .add("error", error.get())
```
jhaber: So you get an `IllegalStateException` if you call `toString` when error isn't set :panda_face:

jonathanwgoodwin: https://github.com/HubSpot/Blazar/blob/slackNotificationsFix/BlazarService/src/main/java/com/hubspot/blazar/externalservice/slack/SlackApiResponse.java#L48
?

PR #141, opened 2016-03-24
Slack ims

jonathanwgoodwin:
@jhaber


BlazarService/src/main/java/com/hubspot/blazar/externalservice/slack/SlackAttachment.java
```diff
@@ -120,4 +121,128 @@ public String getFallback() {
         .add("thumbUrl", thumbUrl)
         .toString();
   }
+
+  public static Builder getBuilder() {
```
jhaber: by convention this would be called `newBuilder`

jhaber:
I think we have enough slack-related stuff now it should get moved to its own repo and we can depend on it in Blazar

jhaber:
cool lgtm

PR #143, opened 2016-03-25
Add a few tests for queued builds

jhaber:
In preparation for allowing multiple queued builds, start adding some tests for the `RepositoryBuildService`. Required implementing `equals`/`hashCode` in a few places and fixing a bug in Rosetta that happens if you use `@StoredAsJson` on an `Optional` field

@jonathanwgoodwin @gchomatas


jhaber:
I'll add some more in a bit, going to merge these while they're passing :dancer:

PR #148, opened 2016-03-30
Support pipe-separated major versions

jhaber:
@jonathanwgoodwin


jhaber:
also @also

jonathanwgoodwin:
I think you'll also need to change these: 
https://github.com/HubSpot/Blazar/blob/master/BlazarService/src/main/java/com/hubspot/blazar/discovery/hsstatic/HubSpotStaticModuleDiscovery.java#L85 to be %s right?

jhaber:
Ah word

PR #151, opened 2016-04-04
make log error more explicit about actions to take

jonathanwgoodwin:
@zdhickman @jhaber @markhazlewood 

Until we're automatically re-trying tasks that are getting `lost` in singularity we should have a better message here so users know what to do. So far I've only see this error come up when a slave has a hung docker daemon.


zdhickman:
:sake:

PR #154, opened 2016-04-08
Kill off repo page in favor of branch selector

zdhickman:
screen shot 2016-04-08 at 11 53 16 am

having trouble with the input autosizer so we're using a fixed width for now...

cc @markhazlewood


jhaber:
:meat_on_bone: :mega: :raised_hands: :cherries:

benrlodge:
🔥 🔥 🔥 

One comment - can we expand the width of the select? It's pretty squished in this example. It looks like it's sizing off the default option, so shorter defaults (eg master) get this affect. Maybe a min-width is an easy workaround.

screenshot 2016-04-12 08 56 12

@zdhickman

zdhickman:
Yeah, good point. I was able to get auto-sizing working correctly, but a min-width would be helpful to have

PR #155, opened 2016-04-12
Reset cache

zdhickman:
screen shot 2016-04-12 at 11 30 54 am

What kind of copy do we want here? How much info/what wording? @markhazlewood @jhaber @jonathanwgoodwin @gchomatas 

Adding copy for the downstream modules tooltip as well

### DIFF
--- a/BlazarUI/app/scripts/components/branch/BranchContainer.jsx
+++ b/BlazarUI/app/scripts/components/branch/BranchContainer.jsx
@@ -2,7 +2,7 @@ import React, {Component, PropTypes} from 'react';
 import {Link} from 'react-router';
 import {Button} from 'react-bootstrap';
 import {bindAll} from 'underscore';
-import classNames from 'classnames'
+import classNames from 'classnames';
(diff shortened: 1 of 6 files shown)


jonathanwgoodwin:
It would be _really_ cool to show what directories were cached last time.
That also nicely covers the case where most HubSpot users don't actually have a use for this feature.

also:
The current copy ("Clear cached directories set in blazar.yaml") is misleading–the cached directories for most builds would come from a buildpack, not a module `.blazar.yaml`.

also:
Oh also, "blazar.yaml" vs. ".blazar.yaml"

zdhickman:
Yep, you might as well replace what I currently have in the tooltip with lorem ipsum

jhaber:
Showing the previous cached directories would be cool but not necessary accurate if the buildpack has changed since the last build

zdhickman:
Okay, maybe we can stick with some succinct copy for now, and add more detail as necessary. Open to suggestions

PR #157, opened 2016-04-13
minimal breadcrumbs

zdhickman:
cc @markhazlewood

### DIFF
--- a/BlazarUI/app/scripts/components/branch/BranchContainer.jsx
+++ b/BlazarUI/app/scripts/components/branch/BranchContainer.jsx
@@ -44,7 +44,7 @@ let initialState = {
   resetCache: false,
   branchId: 0,
   branchInfo: {},
-  branches: [],
+  branchesList: [],
   maxRows: 50
 };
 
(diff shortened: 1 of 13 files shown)


markhazlewood:
I think using `All branches` as the root item is a little confusing. If I'm drilled all the way in to a module's build log, I don't have the name of the repo anywhere.

Maybe try either `[Repo name]` or `[Repo name] branches`

Also the page title is getting redundant:

image

Suggest doing this:

image

And lower down:

image

zdhickman:
I've updated the breadcrumbs to have a more consistent place across the pages it's included in.

I still think it's a bit weird to not have one more level of crumb representing the current page that you're on. It's redundant information, but afaik it's more inline with how breadcrumbs usually work

Here are some screenshots:

screen shot 2016-04-13 at 4 35 25 pm
screen shot 2016-04-13 at 4 35 30 pm
screen shot 2016-04-13 at 4 35 34 pm

also:
🎉  💸

PR #158, opened 2016-04-14
add fullstory tracking info

zdhickman:


also:
Doesn't seem like HubSpot's FullStory configuration belongs here.

zdhickman:
(FYI https://github.com/HubSpot/Blazar/pull/159)

zdhickman:
The whitelisting for stat collection is actually within the Fullstory app. That's unclear in the code, and the org doesn't belong directly in the source anyway.

PR #163, opened 2016-04-15
Allow you configure depends / provides in your .blazar.yaml

jonathanwgoodwin:
@jhaber @gchomatas 
I was working on the tests for inter-project builds and I discovered that though I had seeded the database with module provides & depends data. When I actually triggered a build the modules got re-discovered. Since the config files were `.blazar.yaml` configs they didn't have provides / depends entries so the seeded entries in the db got removed. Which caused my Inter-Project tests to fail.

I'd like to use this to be able to codify those deps in the configs (rather than in the sql migration file).
Also it would let us establish some depends/provides relationships for some of the custom builds we have like `Wailmer Base` > `Build Scripts` > `Wailmer Executor`.


BlazarService/src/main/java/com/hubspot/blazar/util/ModuleBuildLauncher.java
```diff
@@ -108,8 +109,10 @@ private BuildConfig mergeConfig(BuildConfig primary, BuildConfig secondary) {
     Map<String, StepActivationCriteria> stepActivation = new LinkedHashMap<>();
     stepActivation.putAll(secondary.getStepActivation());
     stepActivation.putAll(primary.getStepActivation());
+    Set<String> depends = primary.getDepends().isEmpty() ? secondary.getDepends() : primary.getDepends();
+    Set<String> provides = primary.getProvides().isEmpty() ? secondary.getProvides() : primary.getProvides();
```
jonathanwgoodwin: I didn't merge these because we're currently not merging any lists.
We might want to merge these though because setting up a Java project's dependency on something external would be a huge PITA.

BlazarService/src/main/java/com/hubspot/blazar/util/ModuleBuildLauncher.java
```diff
@@ -109,7 +111,15 @@ private BuildConfig mergeConfig(BuildConfig primary, BuildConfig secondary) {
     stepActivation.putAll(secondary.getStepActivation());
     stepActivation.putAll(primary.getStepActivation());
 
-    return new BuildConfig(steps, before, env, buildDeps, webhooks, cache, Optional.<GitInfo>absent(), Optional.of(user), stepActivation);
+    Set<String> depends = new HashSet<>();
```
jonathanwgoodwin: @jhaber merged per your offline comments.

PR #182, opened 2016-05-04
Checkbox for InterProjectBuilds

jonathanwgoodwin:
- Add checkbox to build modal / hit api to trigger inter-project-build instead of normal one.
- Move state QUEUED -> CALCULATING
- Keep the InterProjectBuildId in the build trigger for builds started by InterProjectBuilds.


jonathanwgoodwin:
@zdhickman we don't need to include 3c87616 in here and it seems d591b02 is an un-finished merge?

PR #183, opened 2016-05-06
Card improvements

zdhickman:
- Column widths: 30%, 20%, 50% instead of 33% all around, to match the expected content size. In the future, maybe this can be even more dynamic
- Trying out adding a matching red background to failing build cards
- Make repo name link to repo page instead of branch page

cc @markhazlewood curious about what you think of the background for the card, or if there is a better way of having failing builds stand out


markhazlewood:
👍  to all of this. Actually fits in well with another future enhancement I've had in mind - "unstar" icon-button visible on hover and aligned right.

PR #185, opened 2016-05-10
Auto trigger

jonathanwgoodwin:
@gchomatas 
This lets us trigger inter project builds on push.
A push to a repository queues a Repository Build with `buildDownstreams` set to INTER_PROJECT.

This Pr introduces `buildSettings` that allow us to control the behavior of a branch for various blazar-related things. I chose to leave slack settings as their own thing, because I felt they should be separate from settings that are blazar-centric even though they will (probably) be changed from the same UI.

When that build figures out what modules it will build we [intercept those modules](https://github.com/HubSpot/Blazar/compare/auto-trigger?expand=1#diff-a0ab931df7a3afd4381dc3f1f7457eccR74) to ensure that your push will only trigger the modules that don't depend on external dependencies (solving the external transitive dependency case) we also start an InterProjectBuild with type "PUSH", and associate (using build mappings) the modules we launch with that build.

After that the inter-project builds happens as normal, modules in the pushed-to repo that were not triggered at the start will get triggered as a child build later on down the graph.

### DIFF
--- a/BlazarBase/src/main/java/com/hubspot/blazar/base/BranchSetting.java
+++ b/BlazarBase/src/main/java/com/hubspot/blazar/base/BranchSetting.java
@@ -7,16 +7,19 @@ public class BranchSetting {
 
   private final long branchId;
   private final boolean triggerInterProjectBuilds;
+  private final boolean interProjectBuildOptIn;
 
   @JsonCreator
   public BranchSetting(@JsonProperty("branchId") long branchId,
-                       @JsonProperty("triggerInterProjectBuilds") boolean triggerInterProjectBuilds) {
+                       @JsonProperty("triggerInterProjectBuilds") boolean triggerInterProjectBuilds,
+                       @JsonProperty("interProjectBuildOptIn") boolean interProjectBuildOptIn) {
     this.branchId = branchId;
     this.triggerInterProjectBuilds = triggerInterProjectBuilds;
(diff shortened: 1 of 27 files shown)


BlazarBase/src/main/java/com/hubspot/blazar/base/BranchSetting.java
```diff
@@ -0,0 +1,36 @@
...
+public class BranchSetting {
+
+  private final long branchId;
+  private final boolean triggerInterProjectBuilds;
+  private final boolean interProjectBuildOptIn;
+
+  @JsonCreator
+  public BranchSetting(@JsonProperty("branchId") long branchId,
+                       @JsonProperty("triggerInterProjectBuilds") boolean triggerInterProjectBuilds,
+                       @JsonProperty("interProjectBuildOptIn") boolean interProjectBuildOptIn) {
```
jonathanwgoodwin: @gchomatas would like some naming input here: 
`triggerInterProjectBuilds` is for auto starting inter project builds when this is triggered
`interProjectBuildOptIn` is for having this (non-master) branch get built as a part of inter project builds.

I've added the second one so that we can slim down the build trees a bit JavaUtils went from ~3300 items to ~650.
This tree now is just master and opted-in branches (hence the setting).

These names are a bit obtuse though, suggestions welcome.

PR #187, opened 2016-05-13
Add UI for branch settings that control inter-project-builds

jonathanwgoodwin:
@zdhickman Needs some styling but the checkboxes work


jonathanwgoodwin:
^ Currently the toggles don't re-render when we make the api call for some reason 👎

jonathanwgoodwin:
Choo choo

PR #188, opened 2016-05-17
Updated styles for branch selector heading.

markhazlewood:
Before:

http://take.ms/jQmG4

After:

http://take.ms/EgiGX


zdhickman:
⛳ 🐐 :godmode:

PR #194, opened 2016-06-01
add focus search bar shortcut

andyhuang91:
@zdhickman @gchomatas @jonathanwgoodwin

### DIFF
--- a/BlazarUI/app/scripts/components/shared/SearchFilter.jsx
+++ b/BlazarUI/app/scripts/components/shared/SearchFilter.jsx
@@ -4,11 +4,13 @@ import Icon from '../shared/Icon.jsx';
 
 import SidebarFilterStore from '../../stores/sidebarFilterStore';
 
+const FOCUS_SEARCH_BAR_SHORTCUTS = [83, 84] // s, t
+
 // To do: Typeahead in input box
 class SearchFilter extends Component {
 
   constructor() {
-    bindAll(this, 'handleChange', 'onStoreChange');
+    bindAll(this, 'handleChange', 'onStoreChange', 'handleKeyup');
     this.state = {
       searchValue: ''
     };
@@ -19,7 +21,7 @@ class SearchFilter extends Component {
       this.props.onChange(this.refs.searchFilterInput.value);
     }, 250);
   }
-    
+
   componentDidMount() {
     window.addEventListener('keyup', this.handleKeyup);
     this.unsubscribeFromSidebarFilter = SidebarFilterStore.listen(this.onStoreChange);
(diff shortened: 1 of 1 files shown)


BlazarUI/app/scripts/components/shared/SearchFilter.jsx
```diff
@@ -19,7 +21,7 @@ class SearchFilter extends Component {
       this.props.onChange(this.refs.searchFilterInput.value);
     }, 250);
   }
-    
+
   componentDidMount() {
     window.addEventListener('keyup', this.handleKeyup);
```
andyhuang91: it looks like this functionality used to be in the UI but was incompletely removed:
https://github.com/HubSpot/Blazar/commit/793d52746bf8e107b6385586007ee14dde840b93

BlazarUI/app/scripts/components/shared/SearchFilter.jsx
```diff
@@ -40,14 +42,23 @@ class SearchFilter extends Component {
...
+  handleChange() {
     this.setState({
       searchValue: this.refs.searchFilterInput.value
     });
     this.handleSearchDebounced();
   }
 
-  render() {    
+  handleKeyup(e) {
+    const notFocusedOnInputElement = e.target === document.body || e.target.tagName === 'A';
```
andyhuang91: it's a little tricky figuring out when this shortcut should be enabled. the tagName part is needed to make sure that the shortcut still works after clicking on a sidebar link.

BlazarUI/app/scripts/components/shared/SearchFilter.jsx
```diff
@@ -40,14 +42,23 @@ class SearchFilter extends Component {
...
+    const modifierKey = e.metaKey || e.shiftKey || event.ctrlKey;
+    const shortcutPressed = FOCUS_SEARCH_BAR_SHORTCUTS.includes(e.which);
```
andyhuang91: Is Array.prototype.includes allowed? It's not supported in IE and babel has a polyfill for it
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes#Browser_compatibility

I can switch it to http://underscorejs.org/#contains if that's safer.

zdhickman: I've been using underscore functions for the sake of consistency, but I don't think it matters much either way (hopefully none of our devs are using IE 🙊)

Maybe just use `contains` to match usage in the rest of the app  -we can swap to `includes` in the future if we want to

jonathanwgoodwin: What Zack said + some external user may at some point try to use IE w/ blazar.

zdhickman:
Beyond the underscore nit, LGTM 👍 thanks!

jonathanwgoodwin:
🎉 🐬 🌊

PR #197, opened 2016-06-07
Inter-project notice

zdhickman:
Ruthlessly simple alert to show information about inter-project builds, and provide links to upstream/downstream repo builds

screen shot 2016-06-07 at 1 45 13 pm
screen shot 2016-06-07 at 1 45 18 pm

/cc @andyhuang91 @jonathanwgoodwin

### DIFF
--- a/BlazarUI/app/scripts/components/repo-build/InterProjectAlert.jsx
+++ b/BlazarUI/app/scripts/components/repo-build/InterProjectAlert.jsx
@@ -1,0 +1,141 @@
+import React, {Component, PropTypes} from 'react';
+import {isEmpty, bindAll} from 'underscore';
+import {Link} from 'react-router';
+import Alert from 'react-bootstrap/lib/Alert';
+
+const initialState = {
+  expanded: false
+};
+
+class InterProjectAlert extends Component {
+
+  constructor(props) {
+    super(props);
+
+    this.state = initialState;
+    bindAll(this, 'onClickAlert');
+  }
+
+  onClickAlert() {
+    this.setState({
+      expanded: !this.state.expanded
+    });
+  }
+
+  renderBuildLinks(repoBuilds) {
+    let renderedLinks = [];
+
+    console.log(repoBuilds);
+
+    Object.keys(repoBuilds).map((value, key) => {
+      renderedLinks.push(
+        <Link key={key} to={`/builds/repo-build/${value}`}>
+          {repoBuilds[value]}
+        </Link>
+      );
+    });
+
+    return renderedLinks;
+  }
+
+  renderUpstreamBuildDetails() {
+    const {upstreamRepoBuilds}  = this.props.upAndDownstreamModules;
+
+    if (Object.keys(upstreamRepoBuilds).length === 0) {
+      return null;
+    }
+
+    return (
+      <div className="upstream-builds">
+        <h4>Upstream builds triggered:</h4>
+        {this.renderBuildLinks(upstreamRepoBuilds)}
+      </div>
+    );
+  }
+
+  renderDownstreamBuildDetails() {
+    const {downstreamRepoBuilds}  = this.props.upAndDownstreamModules;
+
+    if (Object.keys(downstreamRepoBuilds).length === 0) {
+      return null;
+    }
+
+    return (
+      <div className="downstream-builds">
+        <h4>Downstream builds triggered:</h4>
+        {this.renderBuildLinks(downstreamRepoBuilds)}
+      </div>
+    );
+  }
+
+  renderDetails() {
+    return (
+      <div className="inter-project-details">
+        {this.renderUpstreamBuildDetails()}
+        {this.renderDownstreamBuildDetails()}
+      </div>
+    );
+  }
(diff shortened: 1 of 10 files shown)


BlazarUI/app/scripts/components/repo-build/InterProjectAlert.jsx
```diff
@@ -0,0 +1,163 @@
...
+      'inter-project-alert', {
+        'expanded': this.state.expanded
+      }
+    ]);
+  }
+
+  getAlertClass() {
+    const {state} = this.props.upAndDownstreamModules;
+
+    if (contains(['QUEUED', 'IN_PROGRESS'], state)) {
```
andyhuang91: for JS, sometimes people like doing this kind of logic with a map instead of a series of if / case statements

BlazarUI/app/scripts/components/repo-build/InterProjectAlert.jsx
```diff
@@ -0,0 +1,163 @@
...
+
+  renderCancelledModules() {
+    const {cancelledDownstreamModules} = this.props.upAndDownstreamModules;
+
+    if (cancelledDownstreamModules.length === 0) {
+      return null;
+    }
+
+    const renderedCancelledModules = cancelledDownstreamModules.map((module, key) => {
+      console.log(module);
```
zdhickman: :dizzy_face:

BlazarUI/app/scripts/components/repo-build/InterProjectAlert.jsx
```diff
@@ -0,0 +1,163 @@
...
+
+  constructor(props) {
+    super(props);
+
+    this.state = initialState;
+    bindAll(this, 'onClickAlert');
+  }
+
+  getClassNames() {
+    return classNames([
```
andyhuang91: It looks like you don't have to wrap your arguments into an array: `return className('inter-project-alert', {expanded: this.state.expanded})`

BlazarUI/app/scripts/components/repo-build/InterProjectAlert.jsx
```diff
@@ -0,0 +1,163 @@
...
+const initialState = {
+  expanded: false
+};
+
+class InterProjectAlert extends Component {
+
+  constructor(props) {
+    super(props);
+
+    this.state = initialState;
```
andyhuang91: wouldn't this make all components reference the same initial state object?

BlazarUI/app/scripts/components/repo-build/RepoBuildContainer.jsx
```diff
@@ -31,7 +35,9 @@ let initialState = {
   loadingModuleBuilds: true,
   loadingRepoBuild: true,
   loadingStars: true,
-  branchInfo: {}
+  branchInfo: {},
+  upAndDownstreamModules: {},
+  currentRepoBuild: undefined
```
andyhuang91: Can we specifically use `null` when something is purposely not set? `undefined` to me means that you have a object and because of some coding error, it's missing a property that should exist

andyhuang91:
There's a lot of Red in the UI. It kind of gives the impression that all of the downstream builds have some error that I should manually open up and check, but I don't think that's what we're aiming for.

BlazarUI/app/scripts/components/repo-build/InterProjectAlert.jsx
zdhickman: As in, all instances of this alert? Only if this component was not unmounted and we reusing the same instance, which isn't the case

BlazarUI/app/scripts/components/repo-build/RepoBuildContainer.jsx
zdhickman: Definitely

zdhickman:
@jonathanwgoodwin what do you think about that ^

We could change the styling of the inner content, but I think a user who triggers inter-project builds would understand what the status means. So the red doesn't imply that each downstream build is a failure, but instead implies that the status of those builds don't matter in the context of the entire inter-project build

jonathanwgoodwin:
@zdhickman This will show up on builds where the owner did not trigger it. The red would be confusing / overwhelming for someone who didn't expect their repo to get triggered.
Since it shows up on every branch build triggered. I think the blue background you had before with the status text colored might be a better balance.

zdhickman:
Okay, I'll make the status itself colored, and the alert box blue (`info`) styling

BlazarUI/app/scripts/components/repo-build/InterProjectAlert.jsx
andyhuang91: this was a common mistake that people were making in backbone where all instances of a view are referencing the same initial/default object. it probably isn't a big problem since I think all calls to setState sets this.state to a completely new object, but it might lead to subtle bugs. Most of the examples that I've seen online seem to create a new object in the constructor
```
constructor(props) {
  super(props)
  this.state = {
    expanded: false
  }
}

PR #199, opened 2016-06-13
allow you to opt out a git host of github stati

jonathanwgoodwin:
cc: @gchomatas


BlazarService/src/main/java/com/hubspot/blazar/listener/GitHubStatusVisitor.java
```diff
@@ -42,6 +47,11 @@ public void visit(RepositoryBuild build) throws Exception {
     }
 
     GitInfo gitInfo = branchService.get(build.getBranchId()).get();
+    if (!gitHubConfigurationMap.get(gitInfo.getHost()).getSetCommitStatus()) {
```
jonathanwgoodwin: @jhaber I'm not sure I can do this safely.
for our deployment I think this would work, but potentially for others it might not? 
might need some munging of `github.com` to `api.github.com` or the other way around for safety.

PR #203, opened 2016-06-17
WIP: Card stack move

zdhickman:
(This PR is not ready to be merged)

This moves the card stack to its own repo (housed here temporarily: https://github.com/zdhickman/cardstack until we're ready to release the module publicly), which allows us to use card stacks in other react apps.

Some more functionality should be moved into the card stack. It should have memory of the expanded card, and most properties that we are passing into RepoBranchCard that does not relate to the data inside a repo branch card should instead be added inside the card stack package


BlazarUI/app/scripts/components/shared/RepoBranchCard.jsx
```diff
@@ -1,322 +1,13 @@
...
-import ModuleRow from './ModuleRow.jsx';
-import BuildStates from '../../constants/BuildStates';
-import Loader from '../shared/Loader.jsx';
-import Sha from '../shared/Sha.jsx';
+import RepoBranchCardSummary from '../shared/RepoBranchCardSummary.jsx';
+import RepoBranchCardDetails from '../shared/RepoBranchCardDetails.jsx';
 
-class RepoBranchCard extends Card {
+function RepoBranchCard(props) {
+  const RenderedCard = Card(RepoBranchCardSummary, RepoBranchCardDetails);
```
zdhickman: We pass in the summary and details panes as separate components

BlazarUI/app/scripts/components/shared/RepoBranchCardStack.jsx
```diff
@@ -0,0 +1,67 @@
...
+      || this.props.expandedCard !== nextProps.expandedCard
+      || JSON.stringify(this.props.moduleBuildsList) !== JSON.stringify(nextProps.moduleBuildsList);
+  }
+
+  renderCards() {
+    const numberOfBuilds = this.props.starredBuilds.size;
+
+    return this.props.starredBuilds.map((build, key) => {
+      return (
+        <RepoBranchCard
```
zdhickman: onClick, expanded, belowExpanded and loading should be added in by the card stack

BlazarUI/gulpfile.js
```diff
@@ -81,12 +82,15 @@ gulp.task('vendorStyles', function () {
...
+  return gulp.src([
+    app + 'stylus/main.styl'
+  ])
     .pipe($.stylus({
       // only compress if we are in production
       compress: isProduction,
       // include 'normal' css into main.css
-      'include css' : true
+      'include css' : true,
+      include: ['./node_modules/card-stack-test/src/css/card-variables.styl']
```
zdhickman: We can bring in the package's stylus source to be used in our own css

BlazarUI/app/scripts/components/shared/RepoBranchCardStack.jsx
```diff
@@ -0,0 +1,67 @@
...
+          moduleBuildsList={this.props.moduleBuildsList}
+          onClick={() => this.props.onClick(key, build)}
+          key={key}
+          expanded={key === this.props.expandedCard}
+          belowExpanded={key === this.props.expandedCard + 1 && this.props.expandedCard !== -1}
```
andyhuang91: The Card components should only have an expanded property so that we don't have to worry about manually computing and passing in the belowExpanded, first, and last properties. Instead of using the card-stack__card--first, card-stack__card--last, and card-stack__card--below-expanded classes, we could try to use some psuedo classes

BlazarUI/app/scripts/components/shared/RepoBranchCardStack.jsx
```diff
@@ -0,0 +1,67 @@
...
+          item={build}
+          loading={this.props.loading} />
+      );
+    });
+  }
+
+  renderHeader() {
+    return (
+      <RepoBranchCardStackHeader />
```
andyhuang91: If these render methods are going to be so simple, I would probably do them inline instead or assign them to a local variable in the main render function.

BlazarUI/app/scripts/components/shared/RepoBranchCard.jsx
andyhuang91: I'm thinking that it would make more sense to just directly pass in the react element instead of making Card something like a Component factory.
```
const summary = <RepoBranchCardSummary prop1={value1} prop2={value2} />
const details = <RepoBranchCardDetails prop={value}

return <Card summary={summary} details={details} />
```
Card shouldn't need to worry about how to pass in properties to the inner elements.

BlazarUI/app/scripts/components/shared/RepoBranchCardDetails.jsx
```diff
@@ -0,0 +1,127 @@
...
+    });
+  }
+
+  render() {
+    const build = this.getBuildToDisplay();
+
+    let buildTriggerMessage;
+    let detailedTriggerMessage;
+
+    if (build.get('buildTrigger') && build.get('buildTrigger').get('type') === 'MANUAL') {
```
andyhuang91: You can use `build.getIn(['buildTrigger', 'type'])`. It returns undefined if `build.get('buildTrigger')` is not an object

BlazarUI/app/scripts/components/shared/RepoBranchCardDetails.jsx
```diff
@@ -0,0 +1,127 @@
...
+      buildTriggerMessage = 'manually';
+      const buildTime = this.getBuildTime(build.get('startTimestamp'));
+      const author = build.get('buildTrigger').get('id');
```
andyhuang91: Use `getIn` here too

zdhickman:
Made some updates based on comments:

- Card is simplified to accept these four props:

```
<Card
      summary={summary}
      details={details}
      expanded={expanded}
      onClick={onClick}
    />
```

It's hard to move `onClick` as a prop of `CardStack` since we bind the key and build of an individual card to be used later

- No more first last belowExpanded etc
- Brought Loader and ZeroState back out of the card stack as they don't really belong there

BlazarUI/app/scripts/components/shared/RepoBranchCardDetails.jsx
zdhickman: I'll go through and change these all at once where applicable (this is just some copy-pasted code, and that fix can be applied across the app)

andyhuang91:
Will be merged with https://github.com/HubSpot/Blazar/pull/248 instead

PR #206, opened 2016-06-21
add failed builds and inter project build id to alert

zdhickman:
screen shot 2016-06-21 at 5 53 39 pm

---

We're now going to list failed repo builds in the inter project alert box, as well as the id of the inter project build.

@markhazlewood I figure we can do some different styling on the div that contains the failed builds, since they will remain constant even as you navigate through upstream and downstream builds. Maybe we can highlight it in some way

### DIFF
--- a/BlazarUI/app/scripts/components/repo-build/InterProjectAlert.jsx
+++ b/BlazarUI/app/scripts/components/repo-build/InterProjectAlert.jsx
@@ -47,7 +47,7 @@ class InterProjectAlert extends Component {
   }
 
   renderBuildLinks(repoBuilds) {
-    return Object.keys(repoBuilds).map((value, key) => {
+    const listItemNodes = Object.keys(repoBuilds).map((value, key) => {
       return (
         <li key={key}>
           <Link to={`/builds/repo-build/${value}`}>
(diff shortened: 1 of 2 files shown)


jonathanwgoodwin:
cherry-picked into beta here: https://github.com/HubSpot/Blazar/commit/08808e25bccd262862e8cc87af35f9819a4d67b8

zdhickman:
Updated with screenshot

BlazarUI/app/scripts/components/repo-build/InterProjectAlert.jsx
```diff
@@ -56,6 +56,23 @@ class InterProjectAlert extends Component {
...
       );
     });
+
+    return <ul>{listItemNodes}</ul>;
+  }
+
+  renderFailedBuildDetails() {
+    const {failedRepoBuilds} = this.props.upAndDownstreamModules;
+
+    if (Object.keys(failedRepoBuilds).length === 0) {
```
andyhuang91: For future reference, I think most people would just do
```
if (!Object.keys(failedRepoBuilds).length)
```
https://github.com/HubSpot/javascript#comparison--shortcuts

PR #213, opened 2016-06-24
add root build info to interprojectalert

zdhickman:
## Unexpanded
screen shot 2016-07-07 at 2 53 58 pm

---

## Expanded
screen shot 2016-07-07 at 2 54 04 pm

### DIFF
--- a/BlazarUI/app/scripts/components/repo-build/InterProjectAlert.jsx
+++ b/BlazarUI/app/scripts/components/repo-build/InterProjectAlert.jsx
@@ -1,4 +1,5 @@
 import React, {Component, PropTypes} from 'react';
+import ReactDOM from 'react-dom';
 import {isEmpty, bindAll, map} from 'underscore';
 import {Link} from 'react-router';
 import Alert from 'react-bootstrap/lib/Alert';
@@ -44,6 +45,11 @@ class InterProjectAlert extends Component {
     this.setState({
       expanded: !this.state.expanded
     });
+
+    if (this.state.expanded) {
+      const node = ReactDOM.findDOMNode(this);
+      node.scrollTop = 0;
+    }
   }
 
   renderBuildLinks(repoBuilds) {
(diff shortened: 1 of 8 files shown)


jonathanwgoodwin:
@gchomatas

BlazarUI/app/scripts/components/repo-build/InterProjectAlert.jsx
```diff
@@ -162,16 +178,16 @@ class InterProjectAlert extends Component {
       return null;
     }
 
-    const {upstreamRepoBuilds, downstreamRepoBuilds, interProjectBuildId} = upAndDownstreamModules;
+    const {upstreamRepoBuilds, downstreamRepoBuilds, rootRepoBuilds, interProjectBuildId} = upAndDownstreamModules;
 
-    if (Object.keys(upstreamRepoBuilds).length === 0 && Object.keys(downstreamRepoBuilds).length === 0) {
+    if (!(Object.keys(upstreamRepoBuilds).length + Object.keys(downstreamRepoBuilds).length  + Object.keys(rootRepoBuilds).length)) {
```
andyhuang91: I think http://underscorejs.org/#isEmpty represents a little more succinctly what you're trying to do.

andyhuang91:
JS looks okay.

gchomatas:
I don't feel that is a good idea to mix the top-level (root) build(s) within the immediate upstream builds. Even adding the extra info in parenthesis to designate the root build it is confusing.
I would imaging something like that:

Top level (or root) Builds that triggered this build: 
  - build 1
  - buil 2

Immediate Upstream Builds:
 - build 3
 - build 4

Immediate Downstream Builds:
  - build 5
  - build 6

The wording in the titles is just indicative and to make clear what I mean, we can use any other more appropriate / more terse wording 

/cc @markhazlewood

zdhickman:
^ Updated

zdhickman:
Update with new styling

markhazlewood:
@gchomatas I was looking at a different layout/design to better highlight the root in the IPB details that might address your concerns:

image

In the case where you are ON the repo build page that triggered the IPB, the text would just read

`ModuleName [buildNumber]`

Rather than

`Repo::ModuleName [buildNumber]`

markhazlewood:
🚸

PR #214, opened 2016-07-01
Sidebar tweaks

zdhickman:
This partially implements a fix for the sidebar overloading CPU in BlazarUI.

- We poll every 10 seconds instead of every 5
- We don't poll if the tab is not focused

cc @andyhuang91 @gchomatas

### DIFF
--- a/BlazarUI/app/scripts/components/sidebar/SidebarContainer.jsx
+++ b/BlazarUI/app/scripts/components/sidebar/SidebarContainer.jsx
@@ -28,12 +28,14 @@ class SidebarContainer extends Component {
 
   constructor(props) {
     super(props);
-    
-    bindAll(this, 
+
+    bindAll(this,
       'onStoreChange',
-      'onStarChange', 
-      'updateResults', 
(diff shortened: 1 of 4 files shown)


BlazarUI/app/scripts/components/sidebar/SidebarContainer.jsx
```diff
@@ -58,16 +60,28 @@ class SidebarContainer extends Component {
...
+
   componentWillUnmount() {
     BuildsActions.stopPollingBuilds();
     this.unsubscribeFromBuilds();
     window.removeEventListener('resize', this.handleResizeDebounced);
+    window.removeEventListener('blur', this.stopPollingBuilds);
+    window.removeEventListener('focus', this.startPollingBuilds);
+  }
+
+  startPollingBuilds() {
```
zdhickman: I created functions to call these reflux actions so that `add/removeEventListener` had something named to refer to.

andyhuang91:
👍

PR #219, opened 2016-07-19
Use already-polled builds data from "/state" on the repo page

zdhickman:
Now the page (`/builds/repo/<repoName>`) loads literally instantly

cc @gchomatas @andyhuang91


BlazarUI/app/scripts/components/repo/RepoContainer.jsx
```diff
@@ -53,17 +57,18 @@ class RepoContainer extends Component {
     if (state.error) {
       state.loading = false;
     }
+
+    state.builds = state.builds.all;
+
     this.setState(state);
   }
 
   setup(params) {
```
andyhuang91: setup and tearDown don't have to be separate methods now.

BlazarUI/app/scripts/data/StoredBuilds.js
```diff
@@ -1,12 +1,12 @@
 //
 // Super class for polling the builds store
-//  - extended by BranchesApi abnd BranchBuildsApi
+//  - extended by BranchBuildsApi
```
andyhuang91: technically HostsApi extends this too but I don't think it's necessary to mention extended by ... here

BlazarUI/app/scripts/components/repo/RepoContainer.jsx
```diff
@@ -15,11 +15,11 @@ import EmptyMessage from '../shared/EmptyMessage.jsx';
 import {getFilteredBranches, filterInactiveBuilds, sortBranchesByTimestamp} from '../Helpers.js'
 import GenericErrorMessage from '../shared/GenericErrorMessage.jsx';
 
-import RepoStore from '../../stores/repoStore';
-import RepoActions from '../../actions/repoActions';
+import BuildsStore from '../../stores/buildsStore';
+import BuildsActions from '../../actions/buildsActions';
```
andyhuang91: unused?

BlazarUI/app/scripts/components/repo/RepoContainer.jsx
```diff
@@ -36,34 +36,32 @@ class RepoContainer extends Component {
...
   }
 
   componentWillUnmount() {
-    this.tearDown();
+    this.unsubscribeFromBuilds();
   }
 
   onStatusChange(state) {
     if (state.error) {
       state.loading = false;
```
andyhuang91: isn't this bad since you're mutating the passed in object?

BlazarUI/app/scripts/components/repo/RepoContainer.jsx
```diff
@@ -36,34 +36,32 @@ class RepoContainer extends Component {
...
     }
-    this.setState(state);
-  }
 
-  setup(params) {
-    this.unsubscribeFromRepo = RepoStore.listen(this.onStatusChange);
-    RepoActions.loadBranchesAndBuilds(params);
-  }
+    state.builds = state.builds.all;
```
andyhuang91: this component doesn't seem to need all the builds so you can do the filtering here instead of in the render method. I'm assuming that BranchFilter and BranchTable don't need access but it's hard to tell because of how {...this.state}} is passed into both.

andyhuang91:
👍

PR #220, opened 2016-07-19
support fetching branch state host-org-repo

jonathanwgoodwin:
@zdhickman @gchomatas


jonathanwgoodwin:
locally i'm getting sub 300ms for most projects.

jhaber:
Why not just call `getAll()` and filter in Java? The results are already in memory so it should be faster

jonathanwgoodwin:
We've actually decided to leave this for now, since the filtering on the FE has been fixed so it wont wait for 5s randomly.

PR #221, opened 2016-07-20
Some attempt to make slack more reliable and monitor-able

jonathanwgoodwin:
Core issue is that we keep not-sending slack messages and not knowing why. There is an `isOk` method on the reply but as stated [here](https://github.com/Ullink/simple-slack-api/issues/39) this is hard to use / find. 

I've done my best to make it better, but without fixing the 2nd problem (below) we cannot get the error message etc.

The secondary issue is we were depending on the jar for `simple-slack-api` from [mvncentral](http://search.maven.org/#artifactdetails%7Cio.takari.slack%7Csimple-slack-api%7C0.5.1%7Ctakari-jar) BUT that jar isn't really the 0.5.1 jar, because it doesn't have the fixes released in 0.5.1.

cc: @gchomatas @jhaber

### DIFF
--- a/BlazarService/src/main/java/com/hubspot/blazar/listener/SlackImNotificationVisitor.java
+++ b/BlazarService/src/main/java/com/hubspot/blazar/listener/SlackImNotificationVisitor.java
@@ -82,7 +82,10 @@ public class SlackImNotificationVisitor implements RepositoryBuildVisitor {
     }
 
     SlackMessageHandle<SlackMessageReply> result = slackSession.get().sendMessageToUser(user.get(), "", attachment);
-    if (!result.getReply().isOk()) {
+    if (result == null) {
+      LOG.warn("Failed to send slack message to user: {} message: {} slack response was null", email, attachment.toString());
+      return false;
+    } else if (!result.getReply().isOk()) {
       LOG.warn("Failed to send slack message to user: {} message: {} error: {}", email, attachment.toString(), result.getReply().getErrorMessage());
       return false;
     }
--- a/BlazarService/src/main/java/com/hubspot/blazar/listener/SlackImNotificationVisitor.java
+++ b/BlazarService/src/main/java/com/hubspot/blazar/listener/SlackImNotificationVisitor.java
@@ -17,17 +17,15 @@ import com.ullink.slack.simpleslackapi.SlackAttachment;
 import com.ullink.slack.simpleslackapi.SlackMessageHandle;
 import com.ullink.slack.simpleslackapi.SlackSession;
 import com.ullink.slack.simpleslackapi.SlackUser;
-import com.ullink.slack.simpleslackapi.replies.ParsedSlackReply;
 import com.ullink.slack.simpleslackapi.replies.SlackMessageReply;
 
 public class SlackImNotificationVisitor implements RepositoryBuildVisitor {
   private static final Logger LOG = LoggerFactory.getLogger(SlackImNotificationVisitor.class);
   private final BlazarSlackConfiguration blazarSlackConfig;
 
-  private Optional<SlackSession> slackSession;
+  private final Optional<SlackSession> slackSession;
   private final MetricRegistry metricRegistry;
-  private final BlazarConfiguration blazarConfiguration;
-  private SlackUtils slackUtils;
+  private final SlackUtils slackUtils;
 
   @Inject
   public SlackImNotificationVisitor (Optional<SlackSession> slackSession,
@@ -36,7 +34,6 @@ public class SlackImNotificationVisitor implements RepositoryBuildVisitor {
                                      SlackUtils slackUtils) {
     this.slackSession = slackSession;
     this.metricRegistry = metricRegistry;
-    this.blazarConfiguration = blazarConfiguration;
     this.slackUtils = slackUtils;
     this.blazarSlackConfig = blazarConfiguration.getSlackConfiguration().get();
   }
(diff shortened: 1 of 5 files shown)


BlazarService/src/main/java/com/hubspot/blazar/listener/SlackImNotificationVisitor.java
```diff
@@ -50,15 +58,41 @@ public void visit(RepositoryBuild build) throws Exception {
...
-      CommitInfo commitInfo = build.getCommitInfo().get();
-      Optional<SlackUser> user = Optional.fromNullable(slackSession.get().findUserByEmail(commitInfo.getCurrent().getAuthor().getEmail()));
-      if (user.isPresent()) {
-        slackSession.get().sendMessageToUser(user.get(), "", slackUtils.buildSlackAttachment(build));
+    sendSlackMessageWithRetries(build, slackUtils.buildSlackAttachment(build), 3);
+
+  }
+
+  private void sendSlackMessageWithRetries(RepositoryBuild build, SlackAttachment attachment, int tries) {
+    tries = tries - 1;
```
ssalinas: why not use something like Guava retries for all of this? Can set it to retry at varying rates based on return value or specific exceptions a bit easier

BlazarService/src/main/java/com/hubspot/blazar/listener/SlackRoomNotificationVisitor.java
```diff
@@ -147,4 +134,38 @@ private boolean shouldSend(InstantMessageConfiguration instantMessageConfigurati
...
+      if (sendSlackMessage(configuration, attachment)) {
+        return;
+      } else {
+        tries--;
+      }
+    }
+    metricRegistry.counter("failed-slack-channel-sends").inc();
+  }
+
+  private boolean sendSlackMessage(InstantMessageConfiguration configuration, SlackAttachment attachment) {
```
ssalinas: this looks about the same as the implementation in SlackImNotificationVisitor, maybe find a way to not have to duplicate the slack message sending code?

BlazarService/src/main/java/com/hubspot/blazar/util/SlackUtils.java
```diff
@@ -148,4 +153,10 @@ private static SlackAttachment makeAttachment(String title) {
     return new SlackAttachment(title, title, "", "");
   }
 
+  public static Retryer<Boolean> makeSlackMessageSendingRetryer() {
+    return RetryerBuilder.<Boolean>newBuilder()
+        .retryIfResult(Predicates.equalTo(Boolean.FALSE))
+        .withWaitStrategy(WaitStrategies.exponentialWait(3000, TimeUnit.MILLISECONDS))
```
jhaber: I believe we just had someone discover this wait strategy doesn't work as expected

jonathanwgoodwin: Replaced with retry at 1s 3x  times.

BlazarService/src/main/java/com/hubspot/blazar/listener/SlackRoomNotificationVisitor.java
```diff
@@ -31,21 +34,21 @@
   private static final Set<ModuleBuild.State> FAILED_MODULE_STATES = ImmutableSet.of(ModuleBuild.State.CANCELLED, ModuleBuild.State.FAILED);
 
   private InstantMessageConfigurationService instantMessageConfigurationService;
-  private final BranchService branchService;
+  private MetricRegistry metricRegistry;
```
tpetr: forgot a `final` here

BlazarService/src/main/java/com/hubspot/blazar/listener/SlackRoomNotificationVisitor.java
```diff
@@ -60,13 +62,57 @@ public void visit(RepositoryBuild build) throws Exception {
...
+    }
+
+    Optional<SlackChannel> slackChannel = Optional.fromNullable(slackSession.get().findChannelByName(channelName));
+    if (!slackChannel.isPresent()) {
+      LOG.warn("No slack channel found for name {}", channelName);
+      return false;
+    }
+
+    SlackMessageHandle<SlackMessageReply> result = slackSession.get().sendMessage(slackChannel.get(), "", attachment);
+    if (!result.getReply().isOk()) {
```
gchomatas: if (result ==null || !result.getReply().isOk())

if you check the code of the library it is not handling exceptions that gracefully:
```catch (Exception e)
        {
            // TODO : improve exception handling
            e.printStackTrace();
        }```
so you can end up with a null reply

File: BlazarService/src/main/java/com/hubspot/blazar/listener/SlackImNotificationVisitor.java [hunk 6]
```diff
@@ -50,15 +55,37 @@ public void visit(RepositoryBuild build) throws Exception {
...
+    }
+
+    Optional<SlackUser> user = Optional.fromNullable(slackSession.get().findUserByEmail(email));
+    if (!user.isPresent()) {
+      LOG.error("Could not find user {} for message about repo build {}", email, build.getId().get());
+      return false;
+    }
+
+    SlackMessageHandle<SlackMessageReply> result = slackSession.get().sendMessageToUser(user.get(), "", attachment);
+    if (!result.getReply().isOk()) {
```
gchomatas: if (result ==null || !result.getReply().isOk())

PR #222, opened 2016-07-22
Branch state by module no gitInfo included

jonathanwgoodwin:
@andyhuang91 @zdhickman @gchomatas 

I've upgraded moduleState to include the data that we've discussed is required for the new branch-state-by-module page.

The state now includes 2 new builds that it previously didn't have, `lastNonSkippedBuild`, and `lastSuccessfulBuild`. On the FE you should be able to figure out which of the states for the module card you'll be in by comparing those to the already existing `lastBuild`.
You also have `pendingBuild`, `inProgressBuild` to show, when they're present. 

@gchomatas Serialization question
When `pendingBuild` or `inProgressBuild` are not present they're marked as `Optional#absent`, 
But when the `ModuleState` gets serialized I see in the json returned, preferably those aren't there if they're absent, but I'm not sure how to make them go away. (Note: on the absent object the non-optional int/long id fields are actually 0).
```
pendingBuild: {
repoBuildId: 0,
moduleId: 0,
buildNumber: 0
}
```

### DIFF
--- a/BlazarBase/src/main/java/com/hubspot/blazar/base/ModuleState.java
+++ b/BlazarBase/src/main/java/com/hubspot/blazar/base/ModuleState.java
@@ -7,39 +7,39 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.common.base.Optional;
 
 public class ModuleState {
+
   private final Module module;
+  private final Optional<ModuleBuild> lastSuccessfulBuild;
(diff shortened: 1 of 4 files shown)


jonathanwgoodwin:
No endpoint currently also returns a `gitInfo` object and this data, I'm not sure that's needed since you can hit `GET: branch/{id}` to get the info you need there.

This all assumes you have the `branchId`, if not and you need a search by name or something I can build that out.

PR #223, opened 2016-07-25
Use a sql-backed event bus

jhaber:
WIP The issue now is that we do some SQL writes and then enqueue an event to the zookeeper-backed event bus. We do the zookeeper write inside the SQL transaction so that if the enqueue fails the SQL writes will roll back, but the zookeeper write is immediately visible so it can get processed before the SQL transaction completes and those writes become visible (or worse, the SQL transaction could fail to commit after enqueueing the zookeeper event and you're left in an inconsistent state). This should prevent all of those weird edge cases because it will make enqueueing the event part of the transaction, so the writes and the event publish will either all succeed or all fail and they'll all become visible to readers at the same time.

Still need to:
- [x] rework the Guice modules since this doesn't belong in `BlazarZooKeeperModule` anymore
- [x] make sure the webhook only instances can still start up and enqueue events
- [x] add some tests
- [x] update RodanChecks to use the new gauge name

@jonathanwgoodwin @gchomatas


jhaber:
This also adds backoff between retrying queue items (right now they retry immediately) and it permanently fails a queue item after 10 tries (right now they retry forever which has caused it to run out of github api calls in the past or to spam the logs over and over)

gchomatas:
👍  it is the easiest way to have a transactional execution of the the build metadata writing and event publishing

I am curious did mysql ever managed to support two phase commits? :)

jhaber:
It seems like there's some level of support: http://dev.mysql.com/doc/refman/5.7/en/xa.html

PR #224, opened 2016-07-25
Version backfill command

jonathanwgoodwin:
@gchomatas This will pull all of the non-versioned things from the database and iterate over them, running re-discovery on them. 

Kind of concerned about not hammering GH with this one which is why I've added the ratelimit for now, need to tweak that some more to make it more configurable per GH server.

### DIFF
--- a/BlazarBase/src/main/java/com/hubspot/blazar/base/GitInfo.java
+++ b/BlazarBase/src/main/java/com/hubspot/blazar/base/GitInfo.java
@@ -1,13 +1,14 @@
 package com.hubspot.blazar.base;
 
+import java.util.List;
+import java.util.Objects;
+
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.base.MoreObjects;
 import com.google.common.base.Optional;
 import com.google.common.base.Splitter;
-
-import java.util.List;
-import java.util.Objects;
 
 public class GitInfo {
   private final Optional<Integer> id;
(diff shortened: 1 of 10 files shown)


BlazarBase/src/main/java/com/hubspot/blazar/base/GitInfo.java
```diff
@@ -1,14 +1,15 @@
...
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.base.MoreObjects;
 import com.google.common.base.Optional;
 import com.google.common.base.Splitter;
 
-import java.util.List;
-import java.util.Objects;
-
 public class GitInfo {
```
gchomatas: Does this object encapsulate the information for a Git Repository Branch? If yes it could be renamed to GitBranchInfo or GitBranch

BlazarBase/src/main/java/com/hubspot/blazar/base/Module.java
```diff
@@ -97,6 +98,17 @@ public Module withId(int id) {
...
 
   @Override
+  public String toString() {
+    return MoreObjects.toStringHelper(this)
+        .add("id", id)
+        .add("name", name)
+        .add("type", type)
+        .add("path", path)
+        .add("glob", glob)
+        .add("buildpack", buildpack).toString();
```
gchomatas: Why are we naming the info about the git branch this module resides in a buildpack?

BlazarData/src/main/java/com/hubspot/blazar/data/dao/DependenciesDao.java
```diff
@@ -38,6 +39,15 @@
   @SqlQuery("SELECT * FROM module_depends WHERE moduleId = :moduleId")
   Set<Dependency> getDependencies(@Bind("moduleId") int moduleId);
 
+  @SqlQuery("SELECT gitInfo.* " +
```
gchomatas: naming the branch as gitInfo is confusing

jonathanwgoodwin: Yeah this is a big refactor, I think best taken care of in a separate PR given how far it reaches.
We also have to be very careful w/ deployment of that change because the executor isn't deployed at the same time.

BlazarBase/src/main/java/com/hubspot/blazar/base/Module.java
jonathanwgoodwin: Talked offline, answer is: the buildpack object here is a reference to the branch that the buildpack resides on.
A module can have an associated buildpack

PR #226, opened 2016-07-28
Add support for bumping up the resources on your build

jonathanwgoodwin:
This is built off of `0.9.0-hs_stable-SNAPSHOT` when a `0.9.x` release happens I'll move over to that, and will be ready to merge this.

### DIFF
--- a/BlazarBase/src/main/java/com/hubspot/blazar/base/BuildConfig.java
+++ b/BlazarBase/src/main/java/com/hubspot/blazar/base/BuildConfig.java
@@ -9,6 +9,7 @@ import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.common.base.Objects;
 import com.google.common.base.Optional;
+import com.hubspot.mesos.Resources;
 
 public class BuildConfig {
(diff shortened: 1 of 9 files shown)


jonathanwgoodwin:
This has been updated to point to the released Singularity.

BlazarBase/src/main/java/com/hubspot/blazar/base/BuildConfig.java
```diff
@@ -9,6 +9,7 @@
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.common.base.Objects;
 import com.google.common.base.Optional;
+import com.hubspot.mesos.Resources;
```
gchomatas: Let's don't leak model objects from other projects (especially when they are not considered libs) into blazar config.

gchomatas: You may introduce a blazar class to hold the required resources for a build. In this class we can also add some checks to make sure we don't allow users to exceed some limits in requested memory and cpu.

jonathanwgoodwin:
@gchomatas Added a blazar class to represent our view of Resources, and some tests to double check the conversion between the 2 types.

PR #228, opened 2016-08-04
Lint BlazarUI

zdhickman:
@andyhuang91 This PR sets up standard eslint rules, automatically fixes what we can with `--fix`, and fixes scripts in every folder besides `/components`

cc @gchomatas


zdhickman:
All linting errors should be fixed now. There are only linting warnings for `console.warn`s

Calvinp:
You can disable those on a line-by-line basis when you actually want a console warning:
http://eslint.org/docs/user-guide/configuring.html#disabling-rules-with-inline-comments

andyhuang91:
I'm going to say that I trust you but you should probably wait until Monday to release it so that you can quickly fix any problems that people notice.

zdhickman:
I'm going to merge this into `beta` and deploy and wait a few days to see if any hidden bugs come up.

PR #229, opened 2016-08-04
Revert "Add support for bumping up the resources on your build"

jonathanwgoodwin:
Reverts HubSpot/Blazar#226


BlazarService/src/main/java/com/hubspot/blazar/util/SingularityBuildLauncher.java
```diff
@@ -31,18 +29,11 @@ public synchronized void launchBuild(ModuleBuild build) throws Exception {
   private SingularityRunNowRequest buildRequest(ModuleBuild build) {
     String buildId = Long.toString(build.getId().get());
 
-    Optional<com.hubspot.mesos.Resources> mesosResources = Optional.absent();
-    Optional<Resources> blazarResources = build.getBuildConfig().get().getBuildResources();
```
jonathanwgoodwin: This is the line that broke things.
Queued ModuleBuilds don't have a build config _yet_.

PR #231, opened 2016-08-12
Add PostBuildSteps to BuildConfig

jhailman:
This gives build configs the option to specify `after` build actions, similar to the `before` actions:
```
after:
    onSuccess: doCelebrationStuff
    onFailure: doFailureStuff
    always: doCleanupStuff
```
More changes need to be made internally.

cc: @jonathanwgoodwin @gchomatas

### DIFF
--- a/BlazarBase/src/main/java/com/hubspot/blazar/base/BuildConfig.java
+++ b/BlazarBase/src/main/java/com/hubspot/blazar/base/BuildConfig.java
@@ -107,30 +107,4 @@ public class BuildConfig {
     return new BuildConfig(steps, before, after, env, buildDeps, webhooks, cache, buildpack, Optional.of(user), stepActivation, depends, provides);
   }
 
-  public static class PostBuildSteps {
-    private final List<BuildStep> onFailure;
-    private final List<BuildStep> onSuccess;
-    private final List<BuildStep> always;
-
-    @JsonCreator
-    public PostBuildSteps(@JsonProperty("onFailure") List<BuildStep> onFailure,
-                          @JsonProperty("onSuccess") List<BuildStep> onSuccess,
-                          @JsonProperty("always") List<BuildStep> always) {
-      this.onFailure = Objects.firstNonNull(onFailure, Collections.emptyList());
-      this.onSuccess = Objects.firstNonNull(onSuccess, Collections.emptyList());
-      this.always = Objects.firstNonNull(always, Collections.emptyList());
-    }
-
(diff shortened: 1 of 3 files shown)


BlazarService/src/main/java/com/hubspot/blazar/util/ModuleBuildLauncher.java
```diff
@@ -94,6 +94,7 @@ private BuildConfig fetchBuildpack(GitInfo gitInfo) throws IOException, NonRetry
   private BuildConfig mergeConfig(BuildConfig primary, BuildConfig secondary) {
     List<BuildStep> steps = primary.getSteps().isEmpty() ? secondary.getSteps() : primary.getSteps();
     List<BuildStep> before = primary.getBefore().isEmpty() ? secondary.getBefore() : primary.getBefore();
+    Optional<BuildConfig.PostBuildSteps> after = primary.getAfter().isPresent() ? primary.getAfter() : secondary.getAfter();
```
jhailman: How should we merge these two configurations here?

BlazarBase/src/main/java/com/hubspot/blazar/base/BuildConfig.java
```diff
@@ -97,7 +104,33 @@ public static BuildConfig makeDefaultBuildConfig(){
   }
 
   public BuildConfig withUser(String user) {
-    return new BuildConfig(steps, before, env, buildDeps, webhooks, cache, buildpack, Optional.of(user), stepActivation, depends, provides);
+    return new BuildConfig(steps, before, after, env, buildDeps, webhooks, cache, buildpack, Optional.of(user), stepActivation, depends, provides);
   }
 
+  public static class PostBuildSteps {
```
jonathanwgoodwin: This should be its own class.

BlazarService/src/main/java/com/hubspot/blazar/util/ModuleBuildLauncher.java
jonathanwgoodwin: We should use the same strategy as `before` because:
a) We don't allow arbitrarily nested configurations
b) Because of A parent configurations should never set `before` or `after` since they define the core of the build.

SnirkImmington: Noticed my use of `isPresent()` makes that confusing/flips it

jonathanwgoodwin:
LGTM 🚢 
@gchomatas anything else to add?

PR #245, opened 2016-09-16
Hikari + embdedded mysql

jonathanwgoodwin:
This PR attempts to replace the `h2` database we were using before with embedded mysql.

Currently this works for the Data module but something is up with the service module tests.
There are Multiple mysqls being started, which I think is leading to the test queries running against different a one than the schema resulting in errors finding tables in the tests.

CC: @gchomatas


BlazarTestBase/src/main/java/com/hubspot/blazar/test/base/data/HikariDataSourceFactory.java
```diff
@@ -0,0 +1,141 @@
...
+     *  locally and java thinks its in UTC but mysql uses the system timezone which is not UTC.
+     *
+     *  This also avoids having to install and load TimeZone conversion tables to allow mysql
+     *  to convert `UTC` to `+00:00`. If they are installed into the embedded mysql then you
+     *  can simplify the below to:
+     *
+     *  options.put("default-time-zone", Calendar.getInstance().getTimeZone().getID());
+     */
+
+    Duration offset = Duration.ofMillis(Calendar.getInstance().getTimeZone().getRawOffset());
```
jonathanwgoodwin: This part seems a little sketchy to me but I needed this to get timestamp based tests to not fail.

BlazarTestBase/src/main/java/com/hubspot/blazar/test/base/data/HikariDataSourceFactory.java
```diff
@@ -0,0 +1,141 @@
...
+    String jdbcUrl = String.format(jdbcUrlFormat, mysqldResource.getPort(), "blazar");
+    HikariConfig configuration = new HikariConfig();
+    configuration.setJdbcUrl(jdbcUrl);
+    configuration.setDriverClassName("com.mysql.jdbc.Driver");
+    configuration.setUsername("root");
+    configuration.setPassword("");
+
+    HikariManagedDataSource datasource = new HikariManagedDataSource(configuration);
+    try {
+      runSql(datasource, "schema.sql");
```
jonathanwgoodwin: @gchomatas As we discussed this is were we prepare the datasource with the base schema.

BlazarData/src/main/resources/schema.sql
```diff
@@ -1,197 +1,193 @@
 --liquibase formatted sql
```
jonathanwgoodwin: This just reverts back to mysql compliant syntax. H2 did not like using backticks, but not all column names are safe to use without quotes of some kind.

gchomatas: Use quotes only when they are needed and also use a quote type that is compliant with other SQL DBs besides mysql so that other DBs can be potentially used.

BlazarService/src/test/java/com/hubspot/blazar/BlazarServiceTestBase.java
```diff
@@ -3,29 +3,15 @@
...
-      }
-    }
-  }
+public class BlazarServiceTestBase extends DatabaseBackedTest {
 
   @After
-  public void cleanup() throws Exception {
-    runSql("schema.sql");
+  @Inject
+  public void checkEventBusExceptions() throws Exception {
```
gchomatas: Are all services that write in the Db reading/writing in the Event Bus? In other words for which services is this check relevant? Maybe it should be added only in the relevant test classes.

BlazarService/src/test/java/com/hubspot/blazar/service/DependencyVersionTest.java
```diff
@@ -43,11 +45,10 @@
...
   private DependenciesService dependenciesService;
 
+  @Inject
   @Before
-  public void before() throws Exception {
-    // set up the data for these inter-project build tests
-    // todo migrate inter project data to a more common place
-    runSql("InterProjectData.sql");
+  public void before(ManagedDataSource dataSource) throws Exception {
+    runSql(dataSource, "InterProjectData.sql");
```
gchomatas: Shouldn't all migrations done in one place once?

BlazarService/src/test/java/com/hubspot/blazar/service/InterProjectBuildServiceTest.java
```diff
@@ -81,9 +83,10 @@
   private Map<String, GitHub> gitHubMap;
 
   @Before
-  public void before() throws Exception {
+  @Inject
+  public void before(ManagedDataSource dataSource) throws Exception {
     // set up the data for these inter-project build tests
-    runSql("InterProjectData.sql");
+    runSql(dataSource, "InterProjectData.sql");
```
gchomatas: Doing the migration again?

BlazarTestBase/src/main/java/com/hubspot/blazar/test/base/data/HikariDataSourceFactory.java
gchomatas: My understanding is that we convert / use UTC across the database and in all dates processed in the code (or we should do that and let the clients convert the dates/timestamps to local timezones). Why not just simply set mysql to UTC?

BlazarService/src/test/java/com/hubspot/blazar/BlazarServiceTestBase.java
jonathanwgoodwin: All the service layer tests use the event bus.

BlazarService/src/test/java/com/hubspot/blazar/service/DependencyVersionTest.java
jonathanwgoodwin: I should rename this migration, but its just test data, initially added for inter project builds, but now used as a starting point for most service layer tests.

BlazarTestBase/src/main/java/com/hubspot/blazar/test/base/data/HikariDataSourceFactory.java
```diff
@@ -0,0 +1,111 @@
...
+     *  options.put("default-time-zone", Calendar.getInstance().getTimeZone().getID());
+     */
+
+    Duration offset = Duration.ofMillis(Calendar.getInstance().getTimeZone().getRawOffset());
+    long hours = offset.toHours();
+    long minutes = offset.minusHours(hours).toMinutes();
+    // force the + on '+00' for positive offset
+    String mysqlOffset = 0 >= hours ? String.format("+%02d:%02d", hours, minutes) : String.format("%02d:%02d", hours, minutes);
+    options.put("default-time-zone", mysqlOffset);
+    options.put("sql-mode", "ONLY_FULL_GROUP_BY,NO_AUTO_VALUE_ON_ZERO,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION");
```
gchomatas: Tests and the main code should preferably rely on the default mysql settings and not depend on HubSpot specific db settings. Is there a documented, critical sql-mode setting that is a prerequisite for the proper functioning of the code? We may remove the unnecessary options or if we need to keep one document it.

BlazarTestBase/src/main/java/com/hubspot/blazar/test/base/data/HikariDataSourceFactory.java
```diff
@@ -0,0 +1,111 @@
...
+      }
+    }));
+
+    if (!mysqldResource.isRunning()) {
+      throw new RuntimeException("Unable start embedded MySQL");
+    }
+
+    LOG.info("Singleton Embedded MySQL started successfully");
+
+    String jdbcUrlFormat = "jdbc:mysql://127.0.0.1:%d/%s?createDatabaseIfNotExist=true&autoReconnect=true&characterEncoding=UTF8&useSSL=false";
```
gchomatas: In other cases I didn't have to use the useSSL= false option, e.g. Orion doesn't use it and still we don't get ssl warnings. It could be that newer versions of Hikari have ssl as default?

jonathanwgoodwin: Its mysql 5.7 that now messages about it. Does Orion use the mysql5.7 embedded?

BlazarTestBase/src/main/java/com/hubspot/blazar/test/base/data/HikariDataSourceFactory.java
jonathanwgoodwin: Not sure that any of these are "critical" per-se but these settings basically prevent us from accidentally relying on non-obvious default behaviors. `NO_ZERO_IN_DATE` for example throws an error instead of accepting dates like `2012-0-31` 

Almost all of these settings have either become defaults in 5.7 or defaults in 5.7's "strict-mode".
Removing them isn't a huge deal, I'll do that.

PR #247, opened 2016-09-23
Module history endpoint + Fixes to ModuleState query

jonathanwgoodwin:
@gchomatas 

This pr fixes an issue with the ModuleState query where it was returning no results for a module if it did not have both a "last successful build" and a "last non skipped build".

This pr makes a change to the module history endpoint to paginate it. The endpoint returns a `ModuleActivityPage` which holds the results and the number of remaining after that page.


BlazarData/src/main/java/com/hubspot/blazar/data/service/ModuleBuildService.java
```diff
@@ -47,8 +51,32 @@ public ModuleBuildService(ModuleBuildDao moduleBuildDao, ModuleDao moduleDao, Ev
     return moduleBuildDao.getByState(state);
   }
 
-  public List<ModuleBuild> getByModule(int moduleId) {
-    return moduleBuildDao.getByModule(moduleId);
+  public ModuleActivityPage getModuleBuildHistoryPage(int moduleId, Optional<Integer> maybeFromBuildNumber, Optional<Integer> maybePageSize) {
```
andyhuang91: Method name still has 'History' in it instead of activity. Maybe the dao methods should be changed too

BlazarData/src/main/java/com/hubspot/blazar/data/service/ModuleBuildService.java
```diff
@@ -47,8 +51,32 @@ public ModuleBuildService(ModuleBuildDao moduleBuildDao, ModuleDao moduleDao, Ev
...
+      if (maxBuildNumber.isPresent()) {
+        fromBuildNumber = maxBuildNumber.get();
+      } else {
+        return new ModuleActivityPage(ImmutableList.of(), 0);
+      }
+    }
+
+    // get a specific page
+    long start = System.currentTimeMillis();
+    List<ModuleBuildInfo> builds = moduleBuildDao.getLimitedModuleBuildHistory(moduleId, fromBuildNumber, pageSize);
```
andyhuang91: This means we have 3 separate db requests if `fromBuildNumber` is not specified?

BlazarService/src/main/java/com/hubspot/blazar/resources/RepositoryBuildResource.java
```diff
@@ -22,20 +22,24 @@
...
 import com.hubspot.blazar.data.service.ModuleBuildService;
 import com.hubspot.blazar.data.service.RepositoryBuildService;
+import com.hubspot.blazar.data.service.StateService;
 
 @Path("/branches/builds")
 @Produces(MediaType.APPLICATION_JSON)
 public class RepositoryBuildResource {
   private final RepositoryBuildService repositoryBuildService;
   private final ModuleBuildService moduleBuildService;
+  private StateService stateService;
```
andyhuang91: Unused?

BlazarBase/src/main/java/com/hubspot/blazar/base/ModuleState.java
```diff
@@ -11,46 +11,46 @@
 
   private final Module module;
   private final Optional<ModuleBuild> lastSuccessfulModuleBuild;
-  private final Optional<RepositoryBuild> lastSuccessfulRepoBuild;
+  private final Optional<RepositoryBuild> lastSuccessfulBranchBuild;
```
andyhuang91: Would it make sense now to have ModuleState contain ModuleBuildInfos instead of separate fields for module and branch (repo) builds?

BlazarData/src/main/java/com/hubspot/blazar/data/dao/StateDao.java
```diff
@@ -40,39 +41,52 @@
...
+   */
   @SqlQuery("" +
       "SELECT * " +
-      "  FROM modules AS module " +
+      "  FROM branches " +
+      "     LEFT OUTER JOIN modules as module on (branches.id = module.branchId) " +
       // last build
       "     LEFT OUTER JOIN module_builds AS lastModuleBuild ON (module.lastBuildId = lastModuleBuild.id) " +
-      "     LEFT OUTER JOIN repo_builds AS lastRepoBuild ON (lastModuleBuild.repoBuildId = lastRepoBuild.id) " +
+      "     LEFT OUTER JOIN repo_builds AS lastBranchBuild ON (lastModuleBuild.repoBuildId = lastRepoBuild.id) " +
```
andyhuang91: This query is wrong `repo_builds AS lastBranchBuild` but you have `lastRepoBuild.id`

BlazarBase/src/main/java/com/hubspot/blazar/base/ModuleState.java
jonathanwgoodwin: I agree it would make things a bit more standardized, and would probably make sense to return something like that to the FE. I think though I still need this as an intermediate representation, because of how rosetta creates objects from the mysql results.

Would you prefer something like:
```
{'lastSuccessful': { 
  'moduleBuild': <>, 
  'branchBuild': <>
}}
```

BlazarData/src/main/java/com/hubspot/blazar/data/service/ModuleBuildService.java
jonathanwgoodwin: Yeah I'll refactor this

BlazarBase/src/main/java/com/hubspot/blazar/base/ModuleState.java
jonathanwgoodwin: We agreed to do this in a separate PR so we can start User testing this page.

PR #250, opened 2016-09-28
Move a bunch of counter metrics to guages / meters

jonathanwgoodwin:
@gchomatas
This PR adds guages to track the current count of builds at each non-terminal state at both the repo and module level.

I know we discussed tracking the _time_ between Queue - Launching - Running but I think that is best handled separately because it will require a lot of context propagation since currently we have no timestamp information to go off of at any point where that state changes, I'll work on handling that separately.


Jw0x47:
re-making this against hs-master

PR #255, opened 2016-10-05
Log upstream module builds state before launching

jonathanwgoodwin:
A few builds are hanging waiting for upstreams to be complete
before the rest of the build completes, this adds logging to the
code that makes the decision whether to launch the next module.

@gchomatas


gchomatas:
🚢

PR #257, opened 2016-10-05
Redirect beta users to new branch state page

andyhuang91:
For beta users, this will redirect all links to the old branch build history page to the new module centric branch state page. The page now contains a link that will allow users to return to the old page, hosted at a different path, and asks the user for feedback.

screen shot 2016-10-05 at 4 58 16 pm

Currently the flag for whether or not a user is a beta user is determined by looking up a variable in local storage. I'll need to work with @jonathanwgoodwin to determine a way of injecting some code in for our specific environment.

@gchomatas @markhazlewood

### DIFF
--- a/BlazarUI/app/scripts/utils/user.js
+++ b/BlazarUI/app/scripts/utils/user.js
@@ -1,3 +1,5 @@
-import store from 'store';
+import Cookies from 'js-cookie';
 
-export const isBetaUser = store.get('blazar-beta');
+const { betaUserCookie } = window.config;
+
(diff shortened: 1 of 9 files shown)


BlazarUI/app/scripts/utils/user.js
```diff
@@ -0,0 +1,3 @@
+import store from 'store';
+
+export const isBetaUser = store.get('blazar-beta');
```
andyhuang91: @jonathanwgoodwin this is where the current check is. we'll need to inject something here

PR #260, opened 2016-10-05
Better metrics

Jw0x47:
@gchomatas This is ready for review.

I removed the metrics that were increasing counters of number of modules in each state.
I have moved the leader basted metrics inside `LeaderMetricManager` and prefixed all metrics from there with `com.hubspot.blazar.zookeeper.LeaderMetricManager`.

The metrics that will be reported are:

```
# Active builds by state:
com.hubspot.blazar.zookeeper.LeaderMetricManager.ModuleBuild.QUEUED.size
com.hubspot.blazar.zookeeper.LeaderMetricManager.ModuleBuild.WAITING_FOR_UPSTREAM_BUILD.size
com.hubspot.blazar.zookeeper.LeaderMetricManager.ModuleBuild.LAUNCHING.size
com.hubspot.blazar.zookeeper.LeaderMetricManager.ModuleBuild.IN_PROGRESS.size
com.hubspot.blazar.zookeeper.LeaderMetricManager.RepositoryBuild.QUEUED.size
com.hubspot.blazar.zookeeper.LeaderMetricManager.RepositoryBuild.LAUNCHING.size
com.hubspot.blazar.zookeeper.LeaderMetricManager.RepositoryBuild.IN_PROGRESS.size

# Zookeeper queues size by event type
com.hubspot.blazar.zookeeper.LeaderMetricManager.queues.ModuleBuild.size
com.hubspot.blazar.zookeeper.LeaderMetricManager.queues.RepositoryBuild.size
com.hubspot.blazar.zookeeper.LeaderMetricManager.queues.InterProjectBuild.size
com.hubspot.blazar.zookeeper.LeaderMetricManager.queues.DeleteEvent.size
com.hubspot.blazar.zookeeper.LeaderMetricManager.queues.PushEvent.size
com.hubspot.blazar.zookeeper.LeaderMetricManager.queues.CreateEvent.size

# Number of Zk Queues 
com.hubspot.blazar.zookeeper.LeaderMetricManager.queues.size
```

Additionally Dropwizard already reports query times for each query in each dao in the form 
`com.hubspot.blazar.data.dao.DAO_NAME.METHOD_NAME`, I'm currently planning on pulling in that data for the queries we're interested in measuring in the near future.


BlazarData/src/main/java/com/hubspot/blazar/data/dao/MetricsDao.java
```diff
@@ -0,0 +1,19 @@
...
+
+import java.util.Set;
+
+import org.skife.jdbi.v2.sqlobject.SqlQuery;
+
+import com.hubspot.blazar.base.metrics.StateToActiveBuildCountPair;
+
+public interface MetricsDao {
+
+  @SqlQuery("SELECT state AS stateName, COUNT(id) AS count FROM module_builds " +
```
Jw0x47: Explain: 

```
*************************** 1. row ***************************
           id: 1
  select_type: SIMPLE
        table: module_builds
   partitions: NULL
         type: range
possible_keys: state
          key: state
      key_len: 122
          ref: NULL
         rows: 162
     filtered: 100.00
        Extra: Using where; Using index
1 row in set, 1 warning (0.00 sec)
```

BlazarData/src/main/java/com/hubspot/blazar/data/dao/MetricsDao.java
```diff
@@ -0,0 +1,19 @@
...
+      "WHERE state IN ('QUEUED', 'WAITING_FOR_UPSTREAM_BUILD', 'LAUNCHING', 'IN_PROGRESS') GROUP BY state")
+  Set<StateToActiveBuildCountPair> countActiveModuleBuildsByState();
+
+  @SqlQuery("SELECT state AS stateName, COUNT(id) AS count FROM repo_builds " +
```
Jw0x47: Explain (Local after index added)[https://github.com/HubSpot/Blazar/pull/260/files#diff-ec1db078ddc713b22018f83e9ba06d89R195]:

```
*************************** 1. row ***************************
           id: 1
  select_type: SIMPLE
        table: repo_builds
   partitions: NULL
         type: range
possible_keys: build_state
          key: build_state
      key_len: 122
          ref: NULL
         rows: 3
     filtered: 100.00
        Extra: Using where; Using index
1 row in set, 1 warning (0.00 sec)
```

Before index added on prod: 

``````
*************************** 1. row ***************************
           id: 1
  select_type: SIMPLE
        table: repo_builds
   partitions: NULL
         type: ALL
possible_keys: NULL
          key: NULL
      key_len: NULL
          ref: NULL
         rows: 123816
     filtered: 30.00
        Extra: Using where; Using temporary; Using filesort
1 row in set, 1 warning (0.01 sec)```
``````

File: BlazarBase/src/main/java/com/hubspot/blazar/base/metrics/StateToActiveBuildCountPair.java [hunk 3]
```diff
@@ -0,0 +1,45 @@
...
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class StateToActiveBuildCountPair {
+
+  private final String stateName;
+  private final Integer count;
+
+  @JsonCreator
+  public StateToActiveBuildCountPair(@JsonProperty("stateName") String stateName, @JsonProperty("count") Integer count) {
```
gchomatas: make it an int so you spare one check for a null value

BlazarBase/src/main/java/com/hubspot/blazar/base/metrics/StateToActiveBuildCountPair.java
```diff
@@ -0,0 +1,45 @@
...
+    this.stateName = stateName;
```
gchomatas: Check that stateName arg is not null. It doesn't make sense to have a pair with a null state

BlazarBase/src/main/java/com/hubspot/blazar/base/metrics/StateToActiveBuildCountPair.java
```diff
@@ -0,0 +1,45 @@
...
+    if (this == o) {
+      return true;
+    }
+
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+
+    StateToActiveBuildCountPair pair = (StateToActiveBuildCountPair) o;
+    return Objects.equals(stateName, pair.stateName) && Objects.equals(count, pair.count);
```
gchomatas: If no property can be null a plain equals would suffice here

BlazarBase/src/main/java/com/hubspot/blazar/base/metrics/StateToActiveBuildCountPair.java
gchomatas: It would be more clean and easier to communicate the intention if you used a `ModuleStateToActiveBuildCountPair` and a `BranchStateToActiveBuildCountPair` class so that you get rid of the string and use the enumeration for the two states respectively.

BlazarService/src/main/java/com/hubspot/blazar/zookeeper/QueueProcessor.java
```diff
@@ -49,12 +49,12 @@ public QueueProcessor(@Named("QueueProcessor") ScheduledExecutorService executor
                         ZooKeeperEventBus eventBus,
                         ObjectMapper mapper,
                         MetricRegistry metricRegistry,
+                        CachingMetricsService cachingMetricsService,
```
gchomatas: Doesn't seem that CachingMetricsService and MetricRegistry are used in this class

BlazarService/src/main/java/com/hubspot/blazar/zookeeper/LeaderMetricManager.java
```diff
@@ -0,0 +1,108 @@
...
+  private String makeMetricName(Class<?> buildClass, Enum<?> state) {
+    return getClass().getName() + "." + buildClass.getSimpleName() + "." + state.name() + ".size";
+  }
+
+  private Gauge<Number> makeZkQueueGauge(final String path){
+    return () -> {
+      try {
+        return curatorFramework.getChildren().forPath(path).size();
+      } catch (Exception e) {
+        return 0;
```
gchomatas: don't shallow the exception

BlazarData/src/main/java/com/hubspot/blazar/data/service/CachingMetricsService.java
```diff
@@ -0,0 +1,56 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+import com.google.inject.Inject;
+import com.google.inject.Singleton;
+import com.hubspot.blazar.base.ModuleBuild;
+import com.hubspot.blazar.base.RepositoryBuild;
+
+@Singleton
```
gchomatas: It is not a good practice to maintain state in a singleton. This makes your class thread unsafe.
Also in your case you cannot go away by just removing the @Singleton annotation since you expose the instance in metrics registry (through the gauges) which will be potentially used by multiple server threads that will serve requests to the health check endpoint. So you need to synchronize the methods and make the shared properties moduleBuildCountMap, repoBuildCountMap, moduleBuildCountMapLastWrite, repoBuildCountMapLastWrite volatile...or use guava cache that takes care of thread safety.

gchomatas (changes requested):

Jw0x47:
@gchomatas addressed your comments in f11cff6.

BlazarBase/src/main/java/com/hubspot/blazar/base/metrics/StateToActiveBranchBuildCountPair.java
```diff
@@ -0,0 +1,45 @@
+package com.hubspot.blazar.base.metrics;
+
+import java.util.Objects;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.hubspot.blazar.base.RepositoryBuild;
+
+public class StateToActiveBranchBuildCountPair {
```
gchomatas: Just a proposal: 
What about naming the pair classes ActiveBranchBuildsInState, ActiveInterProjectBuildsInState, ActiveModuleBuildsInState respectively? Alternatively we could drop the Pair suffix from the class names.

PR #262, opened 2016-10-06
Allow dismissing beta notification

andyhuang91:
This commit lets the user dismiss the beta notification and saves whether the notification has been dismissed or not in local storage. The branch build history link is moved to the header so that it remains visible even after the notification has been dismissed.

I added a css class to the link so it should be easy for use to see in heap how often people are clicking this button.

screen shot 2016-10-06 at 4 25 35 pm

@gchomatas @jonathanwgoodwin @markhazlewood


gchomatas:
LGTM

PR #246, opened 2016-09-20
Archived repository cleanup job

jonathanwgoodwin:
@gchomatas This command cleans up any active branches that are no longer in organizations blazar manages or on git hosts it no longer manages.

PR #265, opened 2016-10-11
Archived repository cleanup job

Jw0x47:
re-opened https://github.com/HubSpot/Blazar/pull/246 against hs-master.
Cc: @gchomatas


Jw0x47:
@gchomatas I've updated this per your comments in #246

BlazarData/src/test/java/com/hubspot/blazar/data/service/BranchServiceTest.java
```diff
@@ -33,18 +33,20 @@ public void testUpsertBasic() throws Exception {
     assertThat(retrieved.isPresent()).isTrue();
     assertThat(retrieved.get()).isEqualTo(inserted);
     assertThat(retrieved.get().getCreatedTimestamp()).isBetween(before, System.currentTimeMillis());
-    assertThat(retrieved.get().getUpdatedTimestamp()).isBetween(before, System.currentTimeMillis());
-    assertThat(retrieved.get().getUpdatedTimestamp()).isEqualTo(retrieved.get().getCreatedTimestamp());
+
```
gchomatas, line 36: BTW in the above `assertThat(retrieved.get()).isEqualTo(inserted)` assertion you rely on an implementation of equals for GitInfo which is problematic in my opinion. GitInfo is modeling a db entry for a branch adding properties for when this db entry was created or updated. Why then the timestamps are ignored in the implementation of equals. Is it because GitInfo is essentially capturing the info for a branch and we don't care about the db entry specific info in that case? But then why equals is counting the id (surrogate key) that the db entry is getting. 
If you wanted to keep `equals` as it is now it would be better to introduce a BranchInfo class and a separate BranchInfoDbEntry class that adds the surrogate key and timestamp properties.

BlazarData/src/test/java/com/hubspot/blazar/data/service/BranchServiceTest.java
```diff
@@ -20,9 +20,9 @@
   private BranchService branchService;
 
   @Test
-  public void testUpsertBasic() throws Exception {
-    long before = System.currentTimeMillis();
-    Thread.sleep(1000); // because the timestamp is in seconds, we need to ensure we wait at least one.
+  public void itInsertsABranchWhenNonePresent() throws Exception {
+    // Round down to nearest second because mysql is 1s precision
+    long before = Math.floorDiv(System.currentTimeMillis(), 1000L) * 1000;
```
gchomatas: It would suffice to check that getCreatedTimestamp() == getUpdatedTimestamp() and >= System.currentTimeMillis() - 1000 (I mean set `before` to System.currentTimeMillis() - 1000)

Jw0x47: Because the updated timestamp is handled by mysql it is at 1s precision. if the before value is not an even second than the between compare further down will fail, because the value in mysql will be lower than the before due to the precision used.

BlazarData/src/test/java/com/hubspot/blazar/data/service/BranchServiceTest.java
```diff
@@ -33,18 +33,20 @@ public void testUpsertBasic() throws Exception {
...
 
   @Test
-  public void testUpsertRepositoryRename() throws InterruptedException {
+  public void itUpdatesARenamedBranchWhenAlreadyExists() throws InterruptedException {
     long before = System.currentTimeMillis();
     GitInfo original = newGitInfo(123, "Overwatch", "master");
     original = branchService.upsert(original);
-    // The last updated column is `TIMESTAMP` and as such is 1s precision
+    // The last updated column is type `TIMESTAMP` and is 1s precision
+    // We have to wait so that the updated timestamp is actually greater
```
gchomatas: Not worth to delay the test by sleeping just to check the updated timestamp. You can set `before` to System.currentTimeMillis() - 1000

gchomatas:
LGTM let's address my comment on equals at another point

PR #267, opened 2016-10-11
Add a logback appender for sentry to Blazar

Jw0x47:
cc: @gchomatas 

This appender is only used if you configure the logging appender for sentry as follows:

``` yaml
logging:
  appenders:
  - type: sentry
    sentryDsn: <Sentry provided DSN>
    sentryRelease: "QA"
```

### DIFF
--- a/BlazarService/pom.xml
+++ b/BlazarService/pom.xml
@@ -83,6 +83,22 @@
       <groupId>net.sourceforge.argparse4j</groupId>
       <artifactId>argparse4j</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.getsentry.raven</groupId>
+      <artifactId>raven-logback</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>ch.qos.logback</groupId>
+      <artifactId>logback-classic</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>ch.qos.logback</groupId>
+      <artifactId>logback-core</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.dropwizard</groupId>
+      <artifactId>dropwizard-logging</artifactId>
+    </dependency>
     <!-- for jade and jets3t -->
     <dependency>
       <groupId>org.slf4j</groupId>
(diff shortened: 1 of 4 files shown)


BlazarService/src/main/java/com/hubspot/blazar/config/BlazarSentryConfiguration.java
```diff
@@ -0,0 +1,99 @@
...
+ *        sentryDsn: http://some_Sentry_provided_DSN
+ *        sentryRelease: "PROD"
+ * }
+ *
+ */
+@JsonTypeName("sentry")
+public class BlazarSentryConfiguration extends AbstractAppenderFactory<ILoggingEvent> {
+  private static final String APPENDER_NAME = "sentry-appender";
+  private final String sentryDsn;
+  private final String sentryRelease;
```
gchomatas: This could be named `blazarRelease`. `sentryRelease` is confusing because it makes you think that you need to provide the release of sentry.

BlazarService/src/main/java/com/hubspot/blazar/resources/ModuleBuildResource.java
```diff
@@ -54,6 +54,14 @@ public ModuleBuildResource(ModuleBuildService moduleBuildService,
   }
 
   @GET
+  @Path("/break/{value}")
```
gchomatas: I wouldn't leave this in here.

BlazarService/src/main/java/com/hubspot/blazar/config/BlazarSentryConfiguration.java
```diff
@@ -0,0 +1,99 @@
...
+ *  logging:
+ *    appenders:
+ *      - type: sentry
+ *        sentryDsn: http://some_Sentry_provided_DSN
+ *        sentryRelease: "PROD"
+ * }
+ *
+ */
+@JsonTypeName("sentry")
+public class BlazarSentryConfiguration extends AbstractAppenderFactory<ILoggingEvent> {
```
gchomatas: Seems that this class serves as both the appender factory and special dropwizard config class but better name it `SentryAppenderFactory`. It would more easy to understand its main purpose.

gchomatas (changes requested):

PR #278, opened 2016-10-20
Move active and inactive modules to tabs

andyhuang91:
This removes the active / inactive modules headers and replaces them with a tabbed interface. Under the inactive modules tab, there is now some text explaining what inactive modules are.

screen shot 2016-10-20 at 5 30 37 pm


andyhuang91:
Here's a screenshot of the tabs with the new header styling
screen shot 2016-10-21 at 12 59 06 pm

PR #284, opened 2016-11-01
Branch Name validation

Jw0x47:
Adds validation to prevent discovery of branches with weird chars from breaking blazar.
cc: @gchomatas


BlazarService/src/main/java/com/hubspot/blazar/discovery/CompositeModuleDiscovery.java
```diff
@@ -77,4 +85,24 @@ public DiscoveryResult discover(GitInfo gitInfo) throws IOException {
 
     return new DiscoveryResult(modules, malformedFiles);
   }
+
+  /**
+   * There are certain characters which (currently) play havoc with our downstream build tooling (our executor).
```
gchomatas: Let's make more clear what we mean here. Something like: 
We currently don't support branch names with special characters despite the fact that github allows certain special characters in branch names (e.g. a single quote). Having special characters in the branch name causes problems in services that blazar uses to handle builds, i.e. the build executor and the maven service that detects modules.

gchomatas (approved):

PR #286, opened 2016-11-03
Covert git commit info to standard popover

andyhuang91:
Users were confused by the previous 'tooltip' and would try to
clik on it. They were also not aware that they could click
on the links within the previous commit summary tooltip.

Users will now have to click on the link to open the popover
and click anywhere outside of the popover to close it.

screen shot 2016-11-03 at 7 14 21 pm
screen shot 2016-11-03 at 7 14 34 pm
screen shot 2016-11-03 at 7 15 00 pm

### DIFF
--- a/BlazarUI/app/stylus/components/branch-state/commits-summary.styl
+++ b/BlazarUI/app/stylus/components/branch-state/commits-summary.styl
@@ -1,16 +1,21 @@
 @import '../../variables'
 
 .commits-summary
-  color $color-calypso
-  padding 0 5px
+  padding 0 5px !important
+  border none !important
+  font-size inherit
   white-space nowrap
+
+.commits-summary-icon
+  margin-right 5px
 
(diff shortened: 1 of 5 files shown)


BlazarUI/app/stylus/components/branch-state/commits-summary.styl
```diff
@@ -19,42 +24,57 @@
...
   vertical-align middle
   padding-top 0
 
 .commits-summary-poverover__branch-head-icon
   padding-bottom 2px
-  background-color $color-heffalump
+  position relative
+  left 0.5px
+
+  &:before
```
andyhuang91, line 51: I was able to get rid of the branch-head svg and just rely on the octicon and css instead. This should eliminate the brief loading delay that I would sometimes see when triggering the previous tooltip for the first time.

markhazlewood:
🚢

PR #289, opened 2016-11-09
Display an error message for paths with an invalid build number

andyhuang91:
Previously, entering in an invalid build number for the BlazarUI
path /builds/branch/{branchId}/build/{buildNumber} would cause
the app to crash

cc @jonathanwgoodwin

### DIFF
--- a/BlazarUI/app/scripts/data/RepoBuildApi.js
+++ b/BlazarUI/app/scripts/data/RepoBuildApi.js
@@ -12,25 +12,14 @@ function _parse(resp) {
   return resp;
 }
 
-function _fetchModuleNames(params) {
-  const moduleNamesPromise = new Resource({
-    url: `${window.config.apiRoot}/branches/${params.branchId}/modules`,
-    type: 'GET'
-  }).send();
-
-  return moduleNamesPromise;
-}
-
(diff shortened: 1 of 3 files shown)


andyhuang91:
@jonathanwgoodwin Making this change now instead of waiting for api changes to clean up the requests for this page since the error was reported multiple times in Sentry

screen shot 2016-11-09 at 10 45 09 am

BlazarUI/app/scripts/data/RepoBuildApi.js
```diff
@@ -61,20 +61,30 @@ function _getRepoBuildByParam(params, resp) {
   return findWhere(resp, {buildNumber: parseInt(params.buildNumber, 10)});
 }
 
-function fetchRepoBuild(params, cb) {
-  return _fetchBranchBuildHistory(params).then((resp) => {
+function _fetchRepoBuildId(params) {
```
andyhuang91: This is the extra request that we want to eventually get rid of when we update the api. Each time the page polls, it actually makes this request twice just to figure out what repoBuildId corresponds to the build number supplied in the application path.

Jw0x47:
👍 🔥

PR #301, opened 2016-11-29
Filter out modules with no builds on branch state page

andyhuang91:
This prevents a null pointer exception from breaking the page when there
is a module that hasn't been built yet. These modules will be displayed
on the page when we incorporate the styling updates for the module items.

cc @jonathanwgoodwin @markhazlewood


Jw0x47:
@andyhuang91 So that I'm clear, this means that they will just not appear anywhere for now?

andyhuang91:
Yes. By doing that, this fixes the problem with the page not loading at all for these branches. Adding a simple alert didn't look too great. I think it'd be better to have a real design that incorporates unbuilt modules more with the built modules.

Jw0x47:
:shipit:

PR #307, opened 2016-12-01
Fix error handling for builds endpoint

andyhuang91:
The repo page, sidebar, and dashboard all use data from the
same endpoint, which is managed by the BuildsStore.

When an error occurs on the repo page, we now have a message to display
instead of passing the error object into the GenericErrorMessage
component, which results in an exception.

Since errors for the builds endpoint are already shown in the sidebar,
don't display an error message on the repo page if the data has been
previously loaded successfully at least once. Just display whatever
builds have been previously loaded.

If the internet is disconnected, we have status 0 in the response.
The sidebar has been updated to display a more revelant message
instead of "Status 0" and an empty "Response:" section.

Because we are polling the sidebar data, only display an error if we
fail to fetch the data multiple times. This handles the case of a
user's computer taking some time to reconnect to the network more
gracefully.

cc @jonathanwgoodwin @markhazlewood

### DIFF
--- a/BlazarUI/app/scripts/components/repo/RepoContainer.jsx
+++ b/BlazarUI/app/scripts/components/repo/RepoContainer.jsx
@@ -59,16 +59,19 @@ class RepoContainer extends Component {
     });
   }
 
-  onStatusChange(state) {
-    if (state.error) {
-      state.loading = false;
+  onStatusChange(status) {
+    if (status.error && this.state.loading) {
+      this.setState({
+        loading: false,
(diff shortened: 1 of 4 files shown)


andyhuang91:
Instead of this
screen shot 2016-12-01 at 12 17 23 pm
We now display this, and it takes two failed attempts before this message shows up
screen shot 2016-12-01 at 12 18 49 pm

Jw0x47:
🏆 This is awesome work Andy! Thanks for figuring this out.

PR #311, opened 2016-12-05
Update module item styling to new design

andyhuang91:
This new styling improves usability by
1. removing the tabbed interface for the last successful build - users didn't always understand what the last successful tab meant and many users don't care about this information upfront
2. slimming down the module items - users can view more modules at the same time
3. adding a separate build log link in the module item summary - users were confused when they were redirected to a new page instead of expanding the module item when clicking on the module name
4. Making only the module build number link to the build log in the build activity table - users would sometimes unintentionally click on the build activity row, navigating them to the build log page, when trying to interact with the commit summary and relevant user links

screen shot 2016-12-05 at 1 09 24 pm
screen shot 2016-12-05 at 1 09 39 pm

The entire card is still clickable, clicking on the module name now expands the card. The hover state has been keep the same and the module name still has the link color to encourage discoverability that the cards are expandable.

screen shot 2016-12-05 at 1 10 06 pm

The last successful build is now presented only when the card is expanded and there is a tooltip explaining what the number means.
screen shot 2016-12-05 at 1 09 54 pm

The module statuses also now abbreviate the units for build durations at smaller screen sizes to make the page more responsive.
screen shot 2016-12-05 at 12 34 38 pm


Jw0x47:
@andyhuang91 can we toss in an easy win; linking the repo build# in the module grouping to the old repo-build page?

andyhuang91:
@jonathanwgoodwin
Let's discuss tomorrow whether or not we want to add that link. It would provide a slight more streamlined way for people to view the interproject build info, which is currently missing from this page, but we probably shouldn't be encouraging people to use the old page. In addition, we want to eventually integrate interproject build information more directly  into this page in the future. Adding this link would only be a temporary measure and would still be equally unintuitive as the current method now of going through the build history page.

As an additional point of reference, I did try out adding the link and here's how it would look:
screen shot 2016-12-05 at 2 41 48 pm

Visually to me it doesn't stand out much, looks a little awkward with the link color, and could be confusing because the build numbers in the build activity table now links to the individual module build, not the branch build.

Jw0x47:
👍 fyi @gchomatas @markhazlewood

PR #316, opened 2016-12-08
Ungate branch state page

andyhuang91:
This change updates the urls for everyone to link to the new branch state page. There is no more check for beta vs non-beta users. Links have been updated everywhere so the only way to reach the old page is by clicking on the `Build History` link.

I updated the wording a little for the new feature alert. Here's what it currently says
screen shot 2016-12-08 at 5 45 51 pm

### DIFF
--- a/BlazarUI/app/scripts/actions/feedbackActions.js
+++ b/BlazarUI/app/scripts/actions/feedbackActions.js
@@ -2,7 +2,8 @@ import Reflux from 'reflux';
 import Feedback from '../models/Feedback';
 
 const FeedbackActions = Reflux.createActions([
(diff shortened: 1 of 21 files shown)


Jw0x47:
The tone / voice on that message is 👍 🔥 🎉

markhazlewood:
Can you add some emphasis (`<strong>`) on "you'll find the latest build information for each module". 

It will help the most important bit pop out for people who will only be skimming this (which you should assume is **everyone**)

andyhuang91:
Applying mark's updates to bold some of the text and add a link to expand the feedback form from the alert.

screen shot 2016-12-09 at 11 35 21 am

markhazlewood:
⚓️

PR #317, opened 2016-12-08
Fix InterProjectBuild bootstrapping

Jw0x47:
This fixes an issue in InterProjectBuild bootstrapping where projects with a certain structure would not get started properly and a hung-build would result.

cc: @gchomatas


mjball:
zoidberg 🚢 ?

gchomatas:
Looks good, it would be nice if there is a test that emulates the cross repo deps use case that this update fixes.

PR #320, opened 2016-12-20
Auto-expand module item when there is only one module in a branch

andyhuang91:
cc @markhazlewood @jonathanwgoodwin


markhazlewood:
🎄

andyhuang91:
Decided to not go with this functionality after some discussion. We want to discourage users from viewing the build history all the time and instead focus primarily on the current build. We might end up implementing a module page in the future the presents this information directly as we continue move towards a more module centric view of the world / interproject builds.

PR #321, opened 2016-12-20
Moved + resized star button on the branch state page header

markhazlewood:
Moved the star button/icon next to the branch name so that it's clear the branch is being starred and not the whole repo.

### DIFF
--- a/BlazarUI/app/stylus/type.styl
+++ b/BlazarUI/app/stylus/type.styl
@@ -50,6 +50,7 @@ td
 
 .headline-icon
   margin-right 10px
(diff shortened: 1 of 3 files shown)


BlazarUI/app/stylus/type.styl
```diff
@@ -50,6 +50,7 @@ td
 
 .headline-icon
   margin-right 10px
+  margin-left 10px
```
andyhuang91, line 53: Does this affect the headers on the other pages?

andyhuang91:
screenshot?

PR #324, opened 2016-12-29
Make failed and canceled module build history items more visible

andyhuang91:
Adds background color to make the failed/cancelled items stand out more. Users were commenting that the failed items were not visible enough, especially compared to the colored build trigger labels.

screen shot 2016-12-29 at 2 54 17 pm

The divisions between the list items look a bit weird when there are a bunch stacked together
screen shot 2016-12-29 at 2 55 08 pm
screen shot 2016-12-29 at 2 53 56 pm


markhazlewood:
🚢 🚮

PR #325, opened 2017-01-04
Alert users when the last build of a currently in-progress module failed

andyhuang91:
Previously users were only presented with a failing module builds alert
if the current state of a module was failed. This meant that the alert
went away immediately when a new build of the module was triggered.
The alert should disappear only if the module has been successfully built.

*Build in-progress*
screen shot 2017-01-04 at 9 22 36 am

*Build complete with failures*
screen shot 2017-01-04 at 9 23 43 am

cc @markhazlewood @jonathanwgoodwin @gchomatas


andyhuang91:
This is a slightly more awkward case in which the alert mentions build #16 and there is a branch build header section for build #16, but the referenced module build isn't shown there since we only show each module build once

screen shot 2017-01-04 at 10 03 07 am

This is a potential source of confusion for users that we might want to address later on, but I think showing previously failing modules for in-progress modules is definitely a step forward.

markhazlewood:
lgtm  ⚓️

PR #329, opened 2017-01-10
Fix branch state page error after disconnecting from internet

andyhuang91:
Fix occasional `Cannot read property 'get' of undefined` that sometimes appears after a polling request fails

cc @jonathanwgoodwin


Jw0x47:
🚢

PR #332, opened 2017-01-20
Fix extra <td> for debug mode

andyhuang91:
Fixes table row extending beyond table when viewing the branch build page in debug mode

screen shot 2017-01-20 at 1 28 09 pm

cc @jonathanwgoodwin


Jw0x47:
🚢

PR #333, opened 2017-01-20
Fix fonts on dashboard and branch settings pages

andyhuang91:
Previously, fallback fonts were not being used

cc @markhazlewood


Jw0x47:
🚢

PR #335, opened 2017-01-30
Enable slack direct messages + Cleanup.

Jw0x47:
I also refactored the slack code a bunch. While doing that I removed everything that did handled module-build notifications because we decided (a while ago) that it didn't make sense to do notifications at that granularity.

Dms to all users is enabled by making the current `whitelist` empty -- causing the app to fall through to sending to everyone instead of no users. If we find we want to roll back the global direct messages, we can just re-deploy the app with the old whitelist members in it and the notifications will only go to them.


BlazarService/src/main/java/com/hubspot/blazar/guice/BlazarSlackModule.java
```diff
@@ -0,0 +1,77 @@
...
+
+  @Override
+  public void configure(Binder binder) {
+
+    if (!slackConfiguration.isPresent()) {
+      LOG.info("Slack is not configured, not binding slack related resources or slack build notification visitors");
+      return;
+    }
+
+    binder.bind(SlackSession.class).toInstance(makeNewSlackSession());
```
gchomatas: Why don't you create a provider method instead?
Don't you want the session to be a singleton? i.e. keep a single slack session along the app lifecycle? Do you need to get a new session instance each time you inject the session class?

BlazarService/src/main/java/com/hubspot/blazar/guice/BlazarServiceModule.java
```diff
@@ -91,34 +88,33 @@ public void configure(Binder binder) {
     binder.bind(ObjectMapper.class).toInstance(getEnvironment().getObjectMapper());
     Multibinder.newSetBinder(binder, ContainerRequestFilter.class).addBinding().to(GitHubNamingFilter.class).in(Scopes.SINGLETON);
 
+    Multibinder<RepositoryBuildVisitor> repositoryBuildVisitorMultibinder = Multibinder.newSetBinder(binder, RepositoryBuildVisitor.class);
```
gchomatas: Why don't you initialize the Multibinders inside the modules that set their values?

BlazarService/src/main/java/com/hubspot/blazar/resources/InstantMessageResource.java
```diff
@@ -16,32 +12,23 @@
...
-  private static final Logger LOG = LoggerFactory.getLogger(InstantMessageResource.class);
-  @Inject(optional = true)
+  @Inject
   private final InstantMessageConfigurationService instantMessageConfigurationService;
-  private Optional<SlackSession> slackSession;
 
   @Inject
-  public InstantMessageResource(InstantMessageConfigurationService instantMessageConfigurationService,
-                                Optional<SlackSession> slackSession) {
+  public InstantMessageResource(InstantMessageConfigurationService instantMessageConfigurationService) {
```
gchomatas, line 30: Make it more like a restful resource:
@Path("/instant-message-configurations")

Name the class: InstantMessageConfigurationResource()
and then @GET, @POST don't need a @Path (it is @Path("/")) and PUT is @PUT(/{id}), etc.

BlazarService/src/main/java/com/hubspot/blazar/resources/SlackResource.java
```diff
@@ -0,0 +1,38 @@
...
+  private SlackSession slackSession;
+
+  @Inject
+  public SlackResource(SlackSession slackSession) {
+
+    this.slackSession = slackSession;
+  }
+
+  @GET
+  @Path("/list-channels")
```
gchomatas: @Path("/channels")

BlazarService/src/main/java/com/hubspot/blazar/resources/SlackResource.java
```diff
@@ -0,0 +1,38 @@
...
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+
+import com.google.inject.Inject;
+import com.hubspot.blazar.externalservice.slack.SlackChannel;
+import com.ullink.slack.simpleslackapi.SlackSession;
+
+@Path("/instant-message/slack")
```
gchomatas: @Path("/slack")

BlazarService/src/main/java/com/hubspot/blazar/util/SlackClientWrapper.java
```diff
@@ -0,0 +1,129 @@
...
+import com.ullink.slack.simpleslackapi.SlackMessageHandle;
+import com.ullink.slack.simpleslackapi.SlackSession;
+import com.ullink.slack.simpleslackapi.SlackUser;
+import com.ullink.slack.simpleslackapi.replies.ParsedSlackReply;
+import com.ullink.slack.simpleslackapi.replies.SlackMessageReply;
+
+/**
+ * This class handles all the details of sending a slack message, and verifying that it did send.
+ */
+public class SlackClientWrapper {
```
gchomatas: Singleton?

BlazarService/src/main/java/com/hubspot/blazar/util/SlackClientWrapper.java
```diff
@@ -0,0 +1,129 @@
...
+
+      LOG.warn("Failed to send slack message to channel: {} message: {} error: {}", channel.getName(), attachment.toString(), reply.getErrorMessage());
+      return false;
+    }
+    return true;
+  }
+
+
+  private static Retryer<Boolean> makeSlackMessageSendingRetryer() {
+    return RetryerBuilder.<Boolean>newBuilder()
```
gchomatas: The classes that are wrapped by the retrier don't have any try/catch. Are you sure that they are not going to throw any exceptions that your retrier will not handle

BlazarService/src/main/java/com/hubspot/blazar/util/SlackClientWrapper.java
gchomatas: Prefer to use more friendly/less mechanistic names for your classes, i.e. SlackUtils or BlazarSlackClient

BlazarService/src/main/java/com/hubspot/blazar/listener/SlackDmNotificationVisitor.java
```diff
@@ -0,0 +1,91 @@
...
+    this.blazarSlackConfig = blazarConfiguration.getSlackConfiguration().get();
+  }
+
+  @Override
+  public void visit(RepositoryBuild build) throws Exception {
+    Set<String> userEmailsToSendMessagesTo = getOptedInUserEmailsForBuild(build);
+    boolean shouldSendMessage = shouldSendMessage(build, userEmailsToSendMessagesTo);
+    if (shouldSendMessage) {
+      for (String email : userEmailsToSendMessagesTo) {
+        slackClientWrapper.sendMessageToUserByEmail(email, "", slackMessageBuildingUtils.buildSlackAttachment(build));
```
gchomatas: slackClientWrapper.sendMessageToUserByEmail is a bit confusing (sounds like sending an email through slack). The method name could just be `sendMessage` (overloaded) like you have named the private methods (name the private methods `sendMessageWithoutRetries` or something like that)

BlazarService/src/main/java/com/hubspot/blazar/util/SlackMessageBuildingUtils.java
```diff
@@ -0,0 +1,108 @@
...
+import com.google.inject.Inject;
+import com.hubspot.blazar.base.GitInfo;
+import com.hubspot.blazar.base.ModuleBuild;
+import com.hubspot.blazar.base.RepositoryBuild;
+import com.hubspot.blazar.data.service.BranchService;
+import com.hubspot.blazar.data.service.ModuleBuildService;
+import com.hubspot.blazar.data.service.ModuleService;
+import com.ullink.slack.simpleslackapi.SlackAttachment;
+
+public class SlackMessageBuildingUtils {
```
gchomatas: Singleton?

gchomatas (changes requested):

BlazarService/src/main/java/com/hubspot/blazar/guice/BlazarSlackModule.java
Jw0x47: Because we're binding `toInstance` this will not create a new instance each time `SlackSession` is injected, because this is not `Managed` it will not interact with the app lifecycle management that dropwizard does. 

This is from the documentation on `com.google.inject.Binder`:
```
... your module itself, not Guice, takes responsibility for obtaining an instance, then asks Guice to always use this single instance to fulfill all injection requests.
```

I can factor this out to a provides method annotated as Singleton.

BlazarService/src/main/java/com/hubspot/blazar/guice/BlazarServiceModule.java
Jw0x47: Because we add bindings to them in multiple places. It seemed cleaner to me to move all the slack things into `BlazarSlackModule` and then pass in the Multibinders rather than to pass the slack configuration to every class that might need to bind a slack related thing (Visitors, Slack sessions etc.)

BlazarService/src/main/java/com/hubspot/blazar/resources/InstantMessageResource.java
Jw0x47: 👍 Do we want to provide a downtime-free upgrade path for the UI here? If so we need to also provide endpoints that match the old API.

BlazarService/src/main/java/com/hubspot/blazar/util/SlackClientWrapper.java
Jw0x47: Doesn't have to be since it doesn't do anything singleton-y since its using the single `SlackSession` instance. From a clarity point of view would make sense to make this a singleton.

BlazarService/src/main/java/com/hubspot/blazar/resources/InstantMessageResource.java
Jw0x47: cc: @andyhuang91

BlazarService/src/main/java/com/hubspot/blazar/guice/BlazarServiceModule.java
gchomatas: The core idea of the Multibinder is that each module that has to contribute a binding is creating the multibinder and adds its binding. You don't need to pass over the same instance between the modules that add their bindings. The multibinder is registered on a class (or interface) so Guice understands and adds the binding under the same set for a class even if in each module you do Multibinder.newSetBinder() 
From https://github.com/google/guice/wiki/Multibindings
```Note: The method Multibinder.newSetBinder(binder, type) can be confusing. This operation creates a new binder, but doesn't override any existing bindings. A binder created this way contributes to the existing Set of implementations for that type. It would create a new set only if one is not already bound.```

File: BlazarService/src/main/java/com/hubspot/blazar/util/SlackClientWrapper.java (hunk 6 above)
Comment (gchomatas) [reply BlazarService/src/main/java/com/hubspot/blazar/util/SlackClientWrapper.java]: You create a singleton when your class has no state and thus acts like a class of static methods. In this way you don't create a new class each time you inject your singleton class. If your injected class has state then you need a new instance to be injected each time.

**Comment by Jw0x47** [comment]
@gchomatas updated.

File: BlazarService/src/main/java/com/hubspot/blazar/guice/BlazarSlackModule.java [hunk 10]
```diff
@@ -0,0 +1,75 @@
...
+      LOG.info("Slack is not configured, not binding slack related resources or slack build notification visitors");
+      return;
+    }
+
+    Multibinder<RepositoryBuildVisitor> repositoryBuildVisitorMultibinder = Multibinder.newSetBinder(binder, RepositoryBuildVisitor.class);
+    repositoryBuildVisitorMultibinder.addBinding().to(SlackDmNotificationVisitor.class);
+    repositoryBuildVisitorMultibinder.addBinding().to(SlackRoomNotificationVisitor.class);
+
+    binder.bind(SlackSession.class).toInstance(makeSlackSession());
+    binder.bind(UserFeedbackResource.class);
```
gchomatas: you don't need to explicitly bind the following 4 classes since you have the required @Inject and/or @Singleton annotations in the classes.
https://github.com/google/guice/wiki/JustInTimeBindings

BlazarService/src/main/java/com/hubspot/blazar/guice/BlazarSlackModule.java
```diff
@@ -0,0 +1,75 @@
...
+    binder.bind(BlazarSlackClient.class);
+    binder.bind(SlackMessageBuildingUtils.class);
+    binder.bind(SlackResource.class);
+
+    if (slackConfiguration.get().getFeedbackRoom().isPresent()) {
+      binder.bindConstant().annotatedWith(Names.named(SLACK_FEEDBACK_ROOM)).to(slackConfiguration.get().getFeedbackRoom().get());
+      binder.bind(UserFeedbackResource.class);
```
gchomatas: I don't quite understand what you try to do here. You have already bind the `UserFeedBackResource` in line 52 but even if you remove it from there (as I commented above) an implicit binding will happen anyway when the resource class is requested so you can not prevent the binding in this way. 

In any case the best way is to handle the absence of SLACK_FEEDBACK_ROOM in the resource by rejecting the request with a relevant message that will let the client know about the missing channel instead of let the client trying to post at a missing endpoint.

BlazarService/src/main/java/com/hubspot/blazar/guice/BlazarSlackModule.java
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  private SlackSession makeSlackSession() {
+    try {
+      SlackSession session = SlackSessionFactory.createWebSocketSlackSession(slackConfiguration.get().getSlackApiToken());
+      session.connect();
+      return session;
+    } catch (IOException ioe) {
+      LOG.error("Error connecting to Slack", ioe);
+      throw new RuntimeException("Error connecting to Slack", ioe);
```
gchomatas: It is not a good idea to throw an exception when it is not possible to connect to slack. This means we cannot restart blazar and build as long as slack is down. 
 It seems that the client supports connection monitoring and reconnection if it gets past authentication. So if you want to keep it simple you can try to connect and if failed return the session unconnected and delegate to BlazarSlackClient to check if session is connected before it attempts to post.

BlazarService/src/main/java/com/hubspot/blazar/listener/SlackDmNotificationVisitor.java
```diff
@@ -0,0 +1,92 @@
...
+    if (build.getCommitInfo().isPresent()) {
+      LOG.info("No commit info present cannot determine user to slack");
+      return Collections.emptySet();
+    }
+    String authorEmail = build.getCommitInfo().get().getCurrent().getAuthor().getEmail();
+    String committerEmail = build.getCommitInfo().get().getCurrent().getAuthor().getEmail();
+
+    // If the white list is empty we send to whoever pushed
+    // Else we ensure they're also in the whitelist before sending messages
+    if (blazarSlackConfig.getImWhitelist().isEmpty()) {
```
gchomatas: It is better to implement an exclusion list instead. The idea would be to DM everybody unless they are in the exclusion list since we anticipate that all users but a few would like to get a direct message during failures. We can add the exclusion list statically for now and consider maintain it in the UI if there is an interest.

BlazarService/src/main/java/com/hubspot/blazar/listener/SlackDmNotificationVisitor.java
```diff
@@ -0,0 +1,92 @@
...
+    boolean shouldSendMessage = shouldSendMessage(build, userEmailsToSendMessagesTo);
+    if (shouldSendMessage) {
+      for (String email : userEmailsToSendMessagesTo) {
+        String message = "A build started by your code push was not successful";
+        blazarSlackClient.sendMessageToUser(email, message, slackMessageBuildingUtils.buildSlackAttachment(build));
+      }
+    }
+  }
+
+  private Set<String> getOptedInUserEmailsForBuild(RepositoryBuild build) {
```
gchomatas: Maybe rename it to `getUserEmailsToDirectlyNotify`. The users have not really opted in.

BlazarService/src/main/java/com/hubspot/blazar/util/BlazarSlackClient.java
```diff
@@ -0,0 +1,131 @@
...
+        return true;
+      }
+
+      LOG.warn("Failed to send slack message to channel: {} message: {} error: {}", channel.getName(), attachment.toString(), reply.getErrorMessage());
+      return false;
+    }
+    return true;
+  }
+
+  private static Retryer<Boolean> makeSlackMessageSendingRetryer() {
```
gchomatas: Why are you using a static method?

Set the retryer in a `private static final` and reuse the static variable in the methods that you use the retryer

BlazarService/src/main/java/com/hubspot/blazar/util/BlazarSlackClient.java
```diff
@@ -0,0 +1,131 @@
...
+      metricRegistry.counter("successful-slack-dm-sends").inc();
+      // Here we swallow exceptions that might be thrown by our retryer or #sendMessage().
+      // We don't catch and retry any exceptions because the slack session doesn't throw us any, (it swallows them and returns null)
+    } catch (Exception e) {
+      LOG.error("Could not send slack message {}", attachment, e);
+      metricRegistry.counter("failed-slack-dm-sends").inc();
+    }
+  }
+
+  private boolean sendMessage(SlackUser user, String message, SlackAttachment attachment) {
```
gchomatas, line 103: Per my comment above about the session that can be injected not connected, check if session is connected before you try to send

BlazarService/src/main/java/com/hubspot/blazar/util/SlackMessageBuildingUtils.java
```diff
@@ -0,0 +1,110 @@
...
+        String unstableMessage = String.format("Unstable, last attempted build was in branch build #%d", maybePrevious.get().getBuildNumber());
+        attachment.addField(moduleName, unstableMessage, false);
+      } else {
+        attachment.addField(moduleName, "This module has never built.", true);
+      }
+    }
+    return attachment;
+  }
+
+  private String buildNormalStateMessage(RepositoryBuild build) {
```
gchomatas: what do you mean by normal state?

BlazarService/src/main/java/com/hubspot/blazar/util/SlackMessageBuildingUtils.java
```diff
@@ -0,0 +1,110 @@
...
+                                   ModuleBuildService moduleBuildService,
+                                   BlazarUrlHelper blazarUrlHelper) {
+
+    this.branchService = branchService;
+    this.moduleService = moduleService;
+    this.moduleBuildService = moduleBuildService;
+    this.blazarUrlHelper = blazarUrlHelper;
+  }
+
+  public SlackAttachment buildSlackAttachment(RepositoryBuild build) {
```
gchomatas: Add some documentation to explain what it is doing

gchomatas (changes requested):

BlazarService/src/main/java/com/hubspot/blazar/guice/BlazarSlackModule.java
Jw0x47: I've updated us to dropwizard-guicier 1.0.6.0 which supports an option to not enable the Guice Enforcer module which was requiring explicit bindings and have removed the extra bindings here.
I'm going to push another commit that removes any other bindings that are extra.

gchomatas:
🚢

PR #337, opened 2017-02-01
Fix clean metadata

Jw0x47:
@gchomatas 

This fixes issues where the linked blocking queue was not blocking because the CompletableFutures doesn't try to use `submit` when asking for a task to be executed and was just throwing RejectedExecutionExceptions instead.

Also adds a `--noop` flag so that we can run this and see what is going to change.


gchomatas:
🚢

PR #338, opened 2017-02-08
Download url

Jw0x47:
@gchomatas This adds an endpoint to blazar that redirects you to the correct place to download the build log. 

Required for adding the download log button to the build log page.

### DIFF
--- a/BlazarService/src/main/java/com/hubspot/blazar/resources/ModuleBuildResource.java
+++ b/BlazarService/src/main/java/com/hubspot/blazar/resources/ModuleBuildResource.java
@@ -170,7 +170,7 @@ public class ModuleBuildResource {
 
   @GET
   @Path("{id}/log/download")
-  public Response getLog(@PathParam("id") long moduleBuildId) throws URISyntaxException {
+  public Object getLog(@PathParam("id") long moduleBuildId) {
     ModuleBuild build = getBuildWithError(moduleBuildId);
 
     Optional<String> taskId = build.getTaskId();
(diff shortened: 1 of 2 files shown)


BlazarService/src/main/java/com/hubspot/blazar/resources/ModuleBuildResource.java
```diff
@@ -155,6 +168,42 @@ public long getSize() {
     };
   }
 
+  @GET
+  @Path("{id}/log/download")
+  public Response getLog(@PathParam("id") long moduleBuildId) throws URISyntaxException {
```
gchomatas: You should capture the exception and throw a WebException with an appropriate message. It will be confusing for a client (e.g. the UI) getting back a UriSyntaxException.

BlazarService/src/main/java/com/hubspot/blazar/resources/ModuleBuildResource.java
```diff
@@ -155,6 +168,42 @@ public long getSize() {
...
+    Optional<String> taskId = build.getTaskId();
+    if (!taskId.isPresent()) {
+      throw new NotFoundException("No taskId found for build: " + moduleBuildId);
+    }
+
+    Optional<SingularitySandbox> sandboxOptional = singularityClient.browseTaskSandBox(taskId.get(), taskId.get());
+
+    java.util.Optional<SingularitySandboxFile> buildLogFile = java.util.Optional.empty();
+    if (sandboxOptional.isPresent()) {
+      buildLogFile = sandboxOptional.get().getFiles().stream().filter(l -> Objects.equals(l.getName(), BUILD_LOG_NAME)).findFirst();
```
gchomatas: BUILD_LOG_NAME.equals(l.getName())
why are you using Objects?

BlazarService/src/main/java/com/hubspot/blazar/resources/ModuleBuildResource.java
```diff
@@ -155,6 +168,42 @@ public long getSize() {
     };
   }
 
+  @GET
+  @Path("{id}/log/download")
```
gchomatas: @Path("{id}/log/download-url")
public String getBuildLogDownloadUrl
It would be fine to send a redirect once in a while but an endpoint that sends only redirects doesn't seem very clean to me. I think that it is cleaner for the endpoint to be assumed that it generates the log download url. The client can then decide what to do with it. Also the combination of get and /download is ugly and confusing.

BlazarService/src/main/java/com/hubspot/blazar/resources/ModuleBuildResource.java
```diff
@@ -155,6 +168,42 @@ public long getSize() {
...
+      buildLogFile = sandboxOptional.get().getFiles().stream().filter(l -> Objects.equals(l.getName(), BUILD_LOG_NAME)).findFirst();
+    }
+
+    if (buildLogFile.isPresent()) {
+      String host = sandboxOptional.get().getSlaveHostname();
+      int port = singularityConfiguration.getSlaveHttpPort();
+      String path = sandboxOptional.get().getFullPathToRoot() + "/" + sandboxOptional.get().getCurrentDirectory() + "/" + buildLogFile.get().getName();
+      String downloadEndpointPath = "/files/download.json";
+      URI uri = new URI("http", null, host, port, downloadEndpointPath, "?path=" + path, null);
+      return Response.temporaryRedirect(uri).build();
```
gchomatas: Response.ok(uri)

gchomatas (changes requested):

Jw0x47:
@gchomatas @andyhuang91 I've updated this to return an object that contains 1 field `downloadUrl`. 
The plan moving forward is to update the UI to hit this endpoint when someone clicks download and then redirect them to the url returned in the object.
Object looks like this
```
{"downloadUrl":"www.example.com/some-log-file.log"}
```

andyhuang91:
🚢

PR #339, opened 2017-02-14
Fix Inter Project Build commit Resolution

Jw0x47:
If an InterProject build comes just after a push to a repository, but
before the push build is queued it can prevent Blazar from building
the changed modules correctly.

This change makes Blazar use the last built commit for InterProject
builds so that they behave like a "rebuild to pick up dependencies",
instead of having them pick up new commits and not build the right
modules.

### DIFF
--- a/BlazarService/src/main/java/com/hubspot/blazar/listener/GitHubStatusVisitor.java
+++ b/BlazarService/src/main/java/com/hubspot/blazar/listener/GitHubStatusVisitor.java
@@ -2,6 +2,7 @@ package com.hubspot.blazar.listener;
 
 import java.io.FileNotFoundException;
 import java.io.IOException;
+import java.util.List;
 import java.util.Map;
 
 import org.kohsuke.github.GHCommitState;
@@ -9,10 +10,13 @@ import org.kohsuke.github.GHRepository;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.base.Optional;
 import com.google.inject.Inject;
 import com.google.inject.Singleton;
 import com.hubspot.blazar.base.GitInfo;
 import com.hubspot.blazar.base.RepositoryBuild;
+import com.hubspot.blazar.base.github.GitHubApiError;
+import com.hubspot.blazar.base.github.GitHubErrorResponse;
 import com.hubspot.blazar.base.visitor.RepositoryBuildVisitor;
 import com.hubspot.blazar.config.BlazarConfiguration;
 import com.hubspot.blazar.config.GitHubConfiguration;
@@ -23,6 +27,7 @@ import com.hubspot.blazar.util.GitHubHelper;
 @Singleton
 public class GitHubStatusVisitor implements RepositoryBuildVisitor {
   private static final Logger LOG = LoggerFactory.getLogger(GitHubStatusVisitor.class);
+  private static final String GITHUB_TOO_MANY_STATUSES_MESSAGE = "This SHA and context has reached the maximum number of statuses.";
 
   private final BranchService branchService;
   private Map<String, GitHubConfiguration> gitHubConfigurationMap;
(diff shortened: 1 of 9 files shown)


Jw0x47:
CC: @gchomatas

BlazarService/src/main/java/com/hubspot/blazar/util/RepositoryBuildLauncher.java
```diff
@@ -80,7 +88,12 @@ public void launch(RepositoryBuild queued, Optional<RepositoryBuild> previous) t
     }
   }
 
-  private CommitInfo commitInfo(GitInfo gitInfo, Optional<Commit> previousCommit) throws IOException, NonRetryableBuildException {
+  @VisibleForTesting
+  CommitInfo calculateCommitInfoForBuild(
+      GitInfo gitInfo,
+      RepositoryBuild currentBuild,
```
gchomatas: `queuedBuild` to be more accurate?

BlazarService/src/main/java/com/hubspot/blazar/util/RepositoryBuildLauncher.java
```diff
@@ -90,6 +103,15 @@ private CommitInfo commitInfo(GitInfo gitInfo, Optional<Commit> previousCommit)
       throw new NonRetryableBuildException("Couldn't find repository " + gitInfo.getFullRepositoryName(), e);
     }
 
+    Optional<Commit> previousCommit = getCommitFromRepositoryBuild(previousBuild);
```
gchomatas: lastCommitInPreviousBuild

BlazarService/src/main/java/com/hubspot/blazar/util/RepositoryBuildLauncher.java
```diff
@@ -90,6 +103,15 @@ private CommitInfo commitInfo(GitInfo gitInfo, Optional<Commit> previousCommit)
...
+
+    // Inter project builds re-build the project using the commit of the last build if available
+    // otherwise we fall back to using the newest commit available
+    if (currentBuild.getBuildTrigger().getType() == INTER_PROJECT && previousCommit.isPresent()) {
+      return gitHubHelper.commitInfoFor(repository, previousCommit.get(), previousCommit);
```
gchomatas: why don't you just do:
`return new CommitInfo(lastCommitInPreviousBuild, lastCommitInPreviousBuild, ImmutableList.of(), false);`

BlazarService/src/test/java/com/hubspot/blazar/util/RepositoryBuildLauncherTest.java
```diff
@@ -0,0 +1,129 @@
...
+    doAnswer(invocation -> {
+      Commit currentCommit = (Commit) invocation.getArguments()[1];
+      Optional<Commit> previousCommit = (Optional<Commit>) invocation.getArguments()[2];
+      assertThat(currentCommit.getId()).isEqualTo(commitSha);
+      assertThat(previousCommit.isPresent()).isTrue();
+      assertThat(previousCommit.get().getId()).isEqualTo(commitSha);
+      return null;
+    }).when(gitHubHelper).commitInfoFor(any(), any(), any());
+
+    when(gitHubHelper.shaFor(any(), any())).thenThrow(new IllegalStateException("Previous build is present this should not be called"));
```
gchomatas: This kind of indirect testing could be dangerous. If the code changes for any reason your can end up with a test that tests nothing. Prefer to use the assert to directly check the expected outcome of the method that you unit test like `CommitInfo commitInfo = launcher.calculateCommitInfoForBuild()` and then check the commitInfoDirectly. 
As a proof, I already proposed above to directly generate the CommitInfo for IPBs so the `gitHubHelper).commitInfoFor` will not run and the assertions will not happen.

gchomatas (changes requested):

BlazarService/src/main/java/com/hubspot/blazar/listener/GitHubStatusVisitor.java
```diff
@@ -71,6 +76,15 @@ public void visit(RepositoryBuild build) throws Exception {
     try {
       repository.createCommitStatus(sha, state, url, description, "Blazar");
     } catch (IOException e) {
+      Optional<GitHubErrorResponse> gitHubErrorResponse = gitHubHelper.extractErrorResponseFromException(e);
```
Jw0x47, line 79: @gchomatas Refactored things and changed the tests per your suggestions.

The block I'm commenting on here fixes a sentry that we've been getting which is the result of GitHub not letting us post more than 100 statuses. Because we can't delete them, I made this just swallow the exception since there is no point in propagating it if we can't post the status.

BlazarService/src/main/java/com/hubspot/blazar/listener/GitHubStatusVisitor.java
```diff
@@ -71,6 +76,15 @@ public void visit(RepositoryBuild build) throws Exception {
...
+      // GitHub has a limit of 100 statuses you can post to a commit -- since we can't delete statuses we just swallow this error
+      if (gitHubErrorResponse.isPresent()) {
+        List<GitHubApiError> errors = gitHubErrorResponse.get().getErrors();
+        if (errors.stream().filter(er -> er.getMessage().contains(GITHUB_TOO_MANY_STATUSES_MESSAGE)).count() > 0) {
+          LOG.info("Commit {}#{} has too many statuses status posting failed.", gitInfo.getFullRepositoryName(), sha);
```
gchomatas: This is better to be a warning

gchomatas:
🚢

gchomatas:
🚢

PR #340, opened 2017-02-24
Remove public build file

Jw0x47:
Since We're building this internally now; we shouldn't fail the build of the service because of the UI.


andyhuang91:
🚢

PR #344, opened 2017-03-20
Fix issues with namespace clashing in our interal deployer

Jw0x47:
@gchomatas 

Our internal fabric deploy was preventing us from deploying this because there was a config yaml namespace conflict on the key `github`. 

Rather than deal with more config clashes in the future I opted to re-name all the blazar specific options to `blazar####` so that when reading the config it is clear what properties are for the deployer and which are consumed by Blazar.

### DIFF
--- a/BlazarService/src/main/java/com/hubspot/blazar/config/BlazarConfiguration.java
+++ b/BlazarService/src/main/java/com/hubspot/blazar/config/BlazarConfiguration.java
@@ -19,42 +19,60 @@ import io.dropwizard.db.DataSourceFactory;
  */
 public class BlazarConfiguration {
 
+
+  // Defines the deployments of GitHub that Blazar can connect to.
   @NotNull
   @JsonProperty("github")
   private Map<String, GitHubConfiguration> gitHubConfiguration;
 
+  // The configuration required for Blazar to connect to Singularity
   @Valid
   @NotNull
   @JsonProperty("singularity")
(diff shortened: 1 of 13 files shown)


gchomatas:
@jonathanwgoodwin why not just have all `blazar` settings under a `blazar:` property?
i.e. the top level config has only a blazar property which in turn includes all other settings? This seems to be a cleaner and more elegant solution for us and for those that don't mix the dw config with their deploy config.

Jw0x47:
@gchomatas I've updated the PR to nest the blazar configuration inside a single property `blazarConfiguration`.

gchomatas:
🚢

BlazarService/src/main/java/com/hubspot/blazar/config/BlazarConfiguration.java
```diff
@@ -7,52 +7,72 @@
...
 import com.google.common.base.Optional;
 
-import io.dropwizard.Configuration;
 import io.dropwizard.db.DataSourceFactory;
 
-@JsonIgnoreProperties(ignoreUnknown = true)
-public class BlazarConfiguration extends Configuration {
+/**
+ * The configuration for Blazar.
+ * All options that control how the BlazarService behaves are configured here. These are wrapped by {@Link BlazarWrapperConfiguration}
```
aaylward, line 17: s/BlazarWrapperConfiguration/BlazarConfigurationWrapper ?

PR #346, opened 2017-03-23
Fix a bug in slack, bind a visitor that wasnt bound before

Jw0x47:

### DIFF
--- a/BlazarService/src/main/java/com/hubspot/blazar/guice/BlazarServiceModule.java
+++ b/BlazarService/src/main/java/com/hubspot/blazar/guice/BlazarServiceModule.java
@@ -80,10 +80,9 @@ public class BlazarServiceModule extends DropwizardAwareModule<BlazarConfigurati
   @Override
   public void configure(Binder binder) {
     BlazarConfiguration blazarConfiguration = getConfiguration().getBlazarConfiguration();
-    configureBase(binder, blazarConfiguration);
-    configureWebhooks(binder);
+    configureWebhooks(binder, blazarConfiguration);
 
-    // Bind this here so that its available for webhook only instances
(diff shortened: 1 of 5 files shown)


BlazarService/src/main/java/com/hubspot/blazar/guice/BlazarServiceModule.java
```diff
@@ -68,29 +68,29 @@
...
-   * against the same database are then able to process those events using the SQL backed event bus.
-   *
+   * against the same database are then able to process those events using the SQL backed event bus. These
+   * instances are also configured to accept webhooks, but also have the rest of Blazar's API enabled.
    */
   @Override
   public void configure(Binder binder) {
     BlazarConfiguration blazarConfiguration = getConfiguration().getBlazarConfiguration();
-    configureBase(binder, blazarConfiguration);
+    configureWebhooks(binder, blazarConfiguration);
```
Jw0x47, line 83: @gchomatas I re-worked this to work like we discussed offline. To bind all the webhook stuff and then bind extra stuff only if we're not in webhook mode.

BlazarService/src/main/java/com/hubspot/blazar/listener/SlackDmNotificationVisitor.java
```diff
@@ -44,6 +44,10 @@ public SlackDmNotificationVisitor(BlazarConfiguration blazarConfiguration,
 
   @Override
   public void visit(RepositoryBuild build) throws Exception {
+    if (!build.getState().isComplete()) {
```
Jw0x47, line 47: This was required and missing. Otherwise we try to get fields from the RepoBuild that are not filled out because the build is in an early state (Queued etc).

gchomatas:
🚢

PR #341, opened 2017-03-02
Override memory limits

Jw0x47:
Cc: @gchomatas

PR #347, opened 2017-03-24
This should return a default value instead of throwing an exception

Jw0x47:
Was added [here](https://github.com/HubSpot/Blazar/pull/341/files#diff-60d5b71cb20866dcf420d7175db75f4fR161) by mistake.
@gchomatas


gchomatas:
🚢

PR #348, opened 2017-03-24
Reuse branch name

gchomatas:


jhaber:
When a branch is deleted, GitHub sends a deletion event but they also send a push event. In the past, depending on the order the events are processed, this push event would accidentally reactivate the branch. Does this PR handle that case?

BlazarService/src/main/java/com/hubspot/blazar/util/GitHubWebhookHandler.java
```diff
@@ -106,11 +108,7 @@ public void handlePushEvent(PushEvent pushEvent) throws IOException {
 
     if (!pushEvent.getRef().startsWith("refs/tags/") && !pushEvent.getDeleted()) {
```
Jw0x47, line 109: @jhaber The check for `.getDeleted()` here prevents us from re-activating branches that were deleted.

jonathanwgoodwin:
:ship

Jw0x47:
🚢

BlazarService/src/main/java/com/hubspot/blazar/util/GitHubWebhookHandler.java
jhaber: Ah interesting

PR #349, opened 2017-03-27
Updated mysql schema to allow org, repo and branch names to have multibyte UTF chars

gchomatas:
Realized that users use emoticons in their branch names

/cc @jonathanwgoodwin

### DIFF
--- a/BlazarData/src/main/resources/schema.sql
+++ b/BlazarData/src/main/resources/schema.sql
@@ -214,3 +214,8 @@ CREATE TABLE queue_items (
 
 CREATE INDEX build_state_and_timestamps on repo_builds (state, endTimestamp, startTimestamp);
 DROP INDEX build_state on repo_builds;
+
+-- changeset gchomatas:15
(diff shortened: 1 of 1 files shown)


BlazarData/src/main/resources/schema.sql
```diff
@@ -214,3 +214,8 @@ CREATE TABLE queue_items (
 
 CREATE INDEX build_state_and_timestamps on repo_builds (state, endTimestamp, startTimestamp);
 DROP INDEX build_state on repo_builds;
+
+-- changeset gchomatas:15
+ALTER TABLE branches MODIFY `branch` varchar(250) NOT NULL CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
```
Jw0x47, line 219: We've already done this here: https://github.com/HubSpot/Blazar/pull/349/files#diff-ec1db078ddc713b22018f83e9ba06d89R99 I there is something else going on since this did work for a while after we made that change.

Jw0x47:
Correct fix is in: https://github.com/HubSpot/Blazar/pull/351

PR #343, opened 2017-03-17
Fail builds with malformed files or 0 modules.

Jw0x47:
@gchomatas This causes Blazar to fail branch builds for projects with malformed files and builds of branches that have 0 active modules.

PR #350, opened 2017-03-27
Fix a bug in failing malformed module builds

Jw0x47:
@gchomatas Fixes a bug in https://github.com/HubSpot/Blazar/pull/343. The updatePendingBuild needs to be passed a build with the buildId field populated.


gchomatas:
🚢

PR #351, opened 2017-03-28
Mysql considered 🎩 and 💥  to be the same

Jw0x47:
Updating collation to `utf8mb4_bin` fixed the issue.

Mysql's `utf8mb4_unicode_ci` and `utf8mb4_unicode_520_ci` collations can't differentiate emoji correctly. The only way to guarantee mysql correctly identifying uniqueness is to use the binary collation. 

CC: @gchomatas


gchomatas:
🚢

PR #352, opened 2017-03-28
Correctly wrap slack calls with ensureConnected()

Jw0x47:
@gchomatas 

Because we only ran `ensureConnected` inside the `sendMessage*` code calls that tried to look up users / channels that ran before that never ran against a connected slack session. So no messages got sent because slack never connected properly.

This wasn't caught during testing because visiting the SlackConfiguration page in blazar hit a method that did run `ensureConnected()`, which then caused the session to connect. 

I abstracted out the `ensureConnected` code in a way that lets us wrap any call we might make to `slackSession` in this file so that its easier to read the slack code without a bunch of `if (ensureConnected()){}` blocks littered throughout.


gchomatas:
🚢

PR #353, opened 2017-03-28
Update slack room notification visitor

Jw0x47:
Only send channels notifications about builds that built new code.
The notifications for builds that were downstream of other projects
(that did not build new code) generated too many notifications
and was not useful to people.


andyhuang91:
@jonathanwgoodwin Should I make the corresponding change to update the wording in the UI. It looks like we're getting rid of the option to notify on every success (not just ones with commits) so I believe that only a wording change is needed in the UI.

gchomatas:
🚢

PR #357, opened 2017-04-11
Handle singularity outage

Jw0x47:
@gchomatas


Jw0x47:
🚢

BlazarBase/src/main/java/com/hubspot/blazar/base/ModuleBuild.java
```diff
@@ -73,7 +77,8 @@ public ModuleBuild(@JsonProperty("id") Optional<Long> id,
                      @JsonProperty("endTimestamp") Optional<Long> endTimestamp,
                      @JsonProperty("taskId") Optional<String> taskId,
```
andyhuang91, line 78: In the UI I'm using the taskId field directly to generate a link to Singularity when debug mode is enabled. We need to update this class somehow so that I can obtain the correct singularity url.

BlazarData/src/main/java/com/hubspot/blazar/data/service/ModuleBuildService.java
```diff
@@ -115,6 +116,33 @@ public ModuleBuild enqueue(RepositoryBuild repositoryBuild, Module module, Build
   }
 
   @Transactional
+  public ModuleBuild setStateToLaunching(long moduleBuildId, String buildContainerId) {
```
andyhuang91, line 119: this method is named incorrectly

PR #358, opened 2017-04-11
Allow us to silence direct-messages for certain branches

Jw0x47:
I refactored the Direct message configuration into its own class so that its clear what parts of the configuration are for Blazar's slack config in general vs the direct-message part.

@gchomatas


gchomatas:
🚢

PR #360, opened 2017-04-19
better logging for broken upserts

Jw0x47:
We were seeing a sentry on the removed line for a failing `Optional.get()` this will give us a bit more context around what is happening there.


andyhuang91:
🚢

PR #363, opened 2017-05-04
Keep track of which nodes have been cancelled

Jw0x47:
@gchomatas This should stabilize some of our inter-project builds + add some logging that will help us debug in the event of further failures.


gchomatas:
🚢

PR #369, opened 2017-05-10
Add timeout for launching containers

Jw0x47:
Blazar launches the container, and if the cluster says the attempted
launch was successful we wait for 5 minutes for the build executor to tell us
that it has started. If we have not heard from the executor in 5 minutes
we assume that the cluster states NOT_STARTED then we FAIL the build.

If the container state is UNKNOWN then we log an error letting the
maintainer know that Blazar could potentially be leaking containers
into the cluster the build was triggered in.


pranav:
🚢

PR #373, opened 2017-12-18
Remove mathjs dependency

andyhuang91:
GitHub warning about mathjs < 3.17.0 having a high severity security
vulnerability. Since we don't really need to import this library,
removing it instead.

Note: progress.js is legacy code that is not being imported into the
project right now.


tamaccount (approved):

tamaccount:
🚢

